PR #31, opened 2020-09-22
Add service account name to statefulsets

mgianatagh:
Add service account name to statefulsets created by the operator.  This is required for supporting images that require an imagePullSecret.


kminder (approved):
LGTM

PR #36, opened 2020-10-06
Elasticsearch Master should use PVs

pfmackin:
When provisioning an Elasticsearch master, I want to use the minimum size (50g) persistent storage, so that the cluster UUID is saved on persistent storage and doesn't change when the ES pod restarts.  This will prevent the ES data nodes from getting out of sync with the masters (because the UID doesn't change) if all the masters go down at the same time.

Provisioning: 
The ES master uses a Kubernetes stateful set (STS) which provides the ability to automatically create PVC and PV per pod via a volumeClaimTemplate (VTC).   Modify the verrazzano-monitoring-operator to specify the VCT.  Also, change ownership of the ES data directory on the PV so that the ES container can access it.

Deletion:
PVCs are created automatically by Kubernetes when the STS is created because the STS has a volumeClaimTemplate. However, the PVCs are not deleted when the STS is deleted.   In the VMO operator, update the each PVC owner reference to be the StatefulSet (STS) so that when the STS is deleted, the PVC will automatically get deleted. Because PVC is dynamic, when it is deleted, the bound PV will also get deleted. This cannot be done automatically using the STS VolumeClaimTemplate.

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -192,11 +192,12 @@ func CreateZoneAntiAffinityElement(vmoName string, component string) *corev1.Aff
 	}
 }
 
-// GetElasticsearchInitContainerChown return an Elasticsearch Init container which changes owernsip of
-// the ES directory permissions needed to to access PV volume data.  Also set the max map count.
-func GetElasticsearchInitContainerChown() *corev1.Container {
+// GetElasticsearchMasterInitContainer return an Elasticsearch Init container for the mater.  This changes owernship of
+// the ES directory permissions needed to access PV volume data.  Also set the max map count.
+func GetElasticsearchMasterInitContainer() *corev1.Container {
 	elasticsearchInitContainer := CreateContainerElement(nil, nil, config.ElasticsearchInit)
-	elasticsearchInitContainer.Command = []string{"sh", "-c", "chown -R 1000:1000 /usr/share/elasticsearch/data"}
+	elasticsearchInitContainer.Command =
+		[]string{"sh", "-c", "chown -R 1000:1000 /usr/share/elasticsearch/data; sysctl -w vm.max_map_count=262144"}
 	elasticsearchInitContainer.Ports = nil
 	return &elasticsearchInitContainer
 }
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -192,6 +192,15 @@ func CreateZoneAntiAffinityElement(vmoName string, component string) *corev1.Aff
 	}
 }
 
+// GetElasticsearchInitContainerChown return an Elasticsearch Init container which changes owernsip of
+// the ES directory permissions needed to to access PV volume data.  Also set the max map count.
+func GetElasticsearchInitContainerChown() *corev1.Container {
+	elasticsearchInitContainer := CreateContainerElement(nil, nil, config.ElasticsearchInit)
+	elasticsearchInitContainer.Command = []string{"sh", "-c", "chown -R 1000:1000 /usr/share/elasticsearch/data"}
+	elasticsearchInitContainer.Ports = nil
+	return &elasticsearchInitContainer
+}
+
 // GetElasticsearchInitContainer returns an Elasticsearch Init container object
 func GetElasticsearchInitContainer() *corev1.Container {
 	elasticsearchInitContainer := CreateContainerElement(nil, nil, config.ElasticsearchInit)
--- a/pkg/resources/statefulsets/statefulset.go
+++ b/pkg/resources/statefulsets/statefulset.go
@@ -145,7 +145,7 @@ fi`,
 
 	// Add init container
 	statefulSet.Spec.Template.Spec.InitContainers = append(statefulSet.Spec.Template.Spec.InitContainers,
-		*resources.GetElasticsearchInitContainerChown())
+		*resources.GetElasticsearchMasterInitContainer())
 
 	// Add the pv volume mount to the init container
 	statefulSet.Spec.Template.Spec.InitContainers[0].VolumeMounts =
--- a/pkg/resources/statefulsets/statefulset.go
+++ b/pkg/resources/statefulsets/statefulset.go
@@ -5,16 +5,16 @@ package statefulsets
 
 import (
 	"fmt"
-	"strings"
-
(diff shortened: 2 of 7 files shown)


pkg/resources/helper.go
```diff
@@ -192,6 +192,16 @@ func CreateZoneAntiAffinityElement(vmoName string, component string) *corev1.Aff
 	}
 }
 
+// GetElasticsearchMasterInitContainer return an Elasticsearch Init container for the mater.  This changes ownership of
```
bobdonat, line 195: typo "mater"

pkg/vmo/statefulset.go
```diff
@@ -75,3 +88,46 @@ func CreateStatefulSets(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMo
...
+		glog.V(4).Infof("Setting StatefuleSet owner reference for PVC %s", pvc.Name)
+		_, err := controller.kubeclientset.CoreV1().PersistentVolumeClaims(vmoNamespace).Update(context.TODO(), pvc, metav1.UpdateOptions{})
+		if err != nil {
+			glog.Errorf("Failed to update the owner reference in PVC %s, for the reason (%v)", pvc.Name, err)
+			return err
+		}
+	}
+	replicas := int(*statefulSet.Spec.Replicas)
+	numPVCs := len(existingPvcList)
+	if numPVCs != replicas {
```
bobdonat, line 129: How could this happen?

pkg/vmo/statefulset.go
```diff
@@ -75,3 +88,46 @@ func CreateStatefulSets(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMo
 	glog.V(4).Infof("Successfully applied StatefulSets for vmo '%s'", vmo.Name)
 	return nil
 }
+
+// Update the each PVC owner reference to be the StatefulSet (STS).
```
bobdonat, line 92: sentence reads funny

pkg/vmo/statefulset.go
pfmackin: It could be because the pvclister uses a read cache and hasn't been updated with all PVCs that are created asynchronously.  I actually saw this while testing so I added this check.

pkg/resources/helper.go
pfmackin: fixed

pkg/vmo/statefulset.go
pfmackin: fixed

pkg/vmo/statefulset.go
bobdonat: Should the pvclister not be used in this case so it does not rely on the cache and a error is not reported

pfmackin: Actually that should not be an error, i can fix that. In practice it is rarely seen, i tested several times.  However I did see it once.

pfmackin: I think using the lister is ok since this is the normal pattern and is more efficient than calling the API.  Also, I am not positive the lister is the issue since PVCs are created async after the STS is created.

bobdonat (approved):
Looks good to me.

pkg/resources/statefulsets/statefulset.go
```diff
@@ -5,16 +5,16 @@ package statefulsets
...
 	vmcontrollerv1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/config"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/constants"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"strings"
```
mgianatagh, line 17: I am surprised this passes go-fmt, if I re-save this file in IntelliJ the go formatter puts "strings" back on line 8

pfmackin: I had other go fmt errors and explicitly ran go fmt.

pkg/vmo/statefulset.go
```diff
@@ -75,3 +88,46 @@ func CreateStatefulSets(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMo
...
 }
+
+// Update each PVC metadata.ownerReferences field to refer to the StatefulSet (STS).
+// PVCs are created automatically by Kubernetes when the STS is created
+// because the STS has a volumeClaimTemplate.  However, the PVCs are not deleted
+// when the STS is deleted. By setting the PVC metadata.ownerReferences field to refer
+// to the STS resource, the PVC will automatically get deleted when the STS is deleted.
+// Because PVC is dynamic, when it is deleted, the bound PV will also get deleted.
+// NOTE: This cannot be done automatically using the STS VolumeClaimTemplate.
+func updateOwnerForPVCs(controller *Controller, statefulSet *appsv1.StatefulSet, vmoName string, vmoNamespace string) error {
```
mgianatagh, line 99: should statefulset name, UID and replicas be passed by value instead of passing the object?

pkg/resources/statefulsets/statefulset.go
pfmackin: I just moved strings back up top, ran go fmt and it go moved back.  Not sure why different from intellij

bobdonat: I think goimport reorganizes the imports in addition to what gofmt does.  Maybe we should use goimport instead

pkg/vmo/statefulset.go
pfmackin: I think there is a balance between passing a large amount of params and the object.  This is especially true if there are several params of the same type.  Since this is a private helper function specifically for statefulsets I think it is ok.

mgianatagh: ok

mgianatagh (approved):

pkg/resources/statefulsets/statefulset.go
mgianatagh: ok, then this sounds like a potential general issue that is outside the scope of this pull request.

PR #43, opened 2020-10-30
Use env var to check whether to create deployments for app specific v…

ddrechse:
…mi stack

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -5,6 +5,8 @@ package resources
 
 import (
 	"fmt"
+	"github.com/golang/glog"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/resources/helper.go
```diff
@@ -321,3 +323,12 @@ func NewBool(value bool) *bool {
 	var val = value
 	return &val
 }
+
+// IsDevProfile return true if the singleSystemVMI env var is set
+func IsDevProfile(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) bool {
+	_, present := os.LookupEnv("SINGLE_SYSTEM_VMI")
+	glog.V(4).Infof("Env var SINGLE_SYSTEM_VMI present? %v", present)
+	return present
+	//profile := vmo.Spec.Profile
```
mgianatagh, line 332: remove these commented out lines?

ddrechse: Here's the thing with that, that code has already been checked in with another merge and I am trying to limit rebase.  This code will change once I get the new helm chart option configured

mgianatagh (approved):

PR #44, opened 2020-11-04
VZ-1640 - Single-node ES Topology for dekstop/laptop configuration

michael-cico:
Enable a single-node ES configuration for a small machine/workstation configuration

- Single collapsed ES node as master/data/ingest, with ephemeral storage and minimal RAM
- Maintain same services facade for in-cluster clients at least for now
- Triggered by an env-var, SINGLE_SYSTEM_VMI (at least for now)
- Remove es_wait init-container from Kibana pod for now

### DIFF
--- a/pkg/resources/deployments/deployment.go
+++ b/pkg/resources/deployments/deployment.go
@@ -5,6 +5,7 @@ package deployments
 
 import (
 	"fmt"
+	"github.com/golang/glog"
 	"strconv"
 	"strings"
 
@@ -126,7 +127,10 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
 	}
 
 	// Elasticsearch
-	if vmo.Spec.Elasticsearch.Enabled {
+	// - Hacking in the profile for now to skip the ingest/data deployment objects
+	// - We should likely use a factory pattern here to create different VMI profiles, although I think it might
+	//   be better to can the default specs as template files and read them in
+	if vmo.Spec.Elasticsearch.Enabled && !resources.IsDevProfile(vmo) {
 		var es Elasticsearch = ElasticsearchBasic{}
 		deployments = append(deployments, es.createElasticsearchDeploymentElements(vmo, pvcToAdMap)...)
 	}
@@ -152,14 +156,19 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
 		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.PeriodSeconds = 20
 		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.FailureThreshold = 5
 
-		waitForEsInitContainer := corev1.Container{
-			Name:  config.ESWait.Name,
-			Image: config.ESWait.Image,
-			// `-number-of-data-nodes 1` tells eswait to look for at least one data node
-			// `-timeout 5m` tells eswait to wait up to 5 minutes for desired state
-			Args: []string{"-number-of-data-nodes", "1", "-timeout", "5m", elasticsearchURL, config.ESWaitTargetVersion},
+		if !resources.IsDevProfile(vmo) {
+			waitForEsInitContainer := corev1.Container{
+				Name:  config.ESWait.Name,
+				Image: config.ESWait.Image,
+				// `-number-of-data-nodes 1` tells eswait to look for at least one data node
+				// `-timeout 5m` tells eswait to wait up to 5 minutes for desired state
+				Args: []string{"-number-of-data-nodes", "1", "-timeout", "5m", elasticsearchURL, config.ESWaitTargetVersion},
+			}
+			deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, waitForEsInitContainer)
+		} else {
+			glog.Info("In development mode, bypassing ESWait container creation for Kibana")
 		}
-		deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, waitForEsInitContainer)
+
 		deployments = append(deployments, deployment)
 	}
 
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -329,6 +329,4 @@ func IsDevProfile(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) bool {
 	_, present := os.LookupEnv("SINGLE_SYSTEM_VMI")
 	glog.V(4).Infof("Env var SINGLE_SYSTEM_VMI present? %v", present)
 	return present
-	//profile := vmo.Spec.Profile
-	//return len(profile) > 0 && strings.ToLower(profile) == DevelopmentProfile
 }
--- a/pkg/resources/services/elasticsearch.go
+++ b/pkg/resources/services/elasticsearch.go
@@ -6,31 +6,45 @@ package services
 import (
 	vmcontrollerv1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/config"
+	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/constants"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // Creates Elasticsearch Client service element
 func createElasticsearchIngestServiceElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) *corev1.Service {
-	return createServiceElement(vmo, config.ElasticsearchIngest)
+	var elasticsearchIngestService *corev1.Service = createServiceElement(vmo, config.ElasticsearchIngest)
+	if resources.IsDevProfile(vmo) {
+		elasticsearchIngestService.Spec.Selector = resources.GetSpecID(vmo.Name, config.ElasticsearchMaster.Name)
+		// In dev mode, only a single node/pod all ingest/data goes to the 9200 port on the back end node
+		elasticsearchIngestService.Spec.Ports = []corev1.ServicePort{resources.GetServicePort(config.ElasticsearchData)}
+	}
+	return elasticsearchIngestService
 }
 
 // Creates Elasticsearch Master service element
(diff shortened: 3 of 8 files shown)


pkg/resources/statefulsets/statefulset.go
```diff
@@ -36,9 +38,16 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) ([]*appsv1.StatefulSe
...
 	var readinessProbeCondition string
 
+	devProfile := resources.IsDevProfile(vmo)
+	glog.Info("devProfile: ", strconv.FormatBool(devProfile))
+
 	statefulSet := createStatefulSetElement(vmo, &vmo.Spec.Elasticsearch.MasterNode.Resources, config.ElasticsearchMaster, "")
 
-	statefulSet.Spec.Replicas = resources.NewVal(vmo.Spec.Elasticsearch.MasterNode.Replicas)
+	if devProfile {
+		statefulSet.Spec.Replicas = resources.NewVal(1)
```
ddrechse, line 47: Maybe the 1 could be a const

michael-cico: Fixed

pkg/resources/deployments/deployment.go
```diff
@@ -152,14 +156,19 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
 		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.PeriodSeconds = 20
 		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.FailureThreshold = 5
 
-		waitForEsInitContainer := corev1.Container{
-			Name:  config.ESWait.Name,
-			Image: config.ESWait.Image,
-			// `-number-of-data-nodes 1` tells eswait to look for at least one data node
-			// `-timeout 5m` tells eswait to wait up to 5 minutes for desired state
-			Args: []string{"-number-of-data-nodes", "1", "-timeout", "5m", elasticsearchURL, config.ESWaitTargetVersion},
+		if !resources.IsDevProfile(vmo) {
```
pfmackin, line 159: Doesn't Kibana still have to wait for ES?

pkg/resources/helper.go
```diff
@@ -321,3 +323,12 @@ func NewBool(value bool) *bool {
 	var val = value
 	return &val
 }
+
+// IsDevProfile return true if the singleSystemVMI env var is set
+func IsDevProfile(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) bool {
+	_, present := os.LookupEnv("SINGLE_SYSTEM_VMI")
+	glog.Infof("CDD Env var SINGLE_SYSTEM_VMI present? %v", present)
```
bobdonat, line 330: what is CDD?

pkg/resources/helper.go
```diff
@@ -321,3 +323,12 @@ func NewBool(value bool) *bool {
...
+	return present
+	//profile := vmo.Spec.Profile
```
bobdonat, line 332: should commented out code be removed

pkg/resources/helper.go
```diff
@@ -321,3 +323,12 @@ func NewBool(value bool) *bool {
 	var val = value
 	return &val
 }
+
+// IsDevProfile return true if the singleSystemVMI env var is set
+func IsDevProfile(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) bool {
+	_, present := os.LookupEnv("SINGLE_SYSTEM_VMI")
```
pfmackin, line 329: SINGLE_SYSTEM_VMI is sort of confusing.   We already a single system VMI.  Maybe DEV_VMI would be better

pkg/resources/deployments/deployment.go
michael-cico: I removed that dependency for now, based on @gilbode 's input.  We may have to revisit, but likely its OK for now.  I should add that to the PR description though.

pkg/resources/services/elasticsearch.go
```diff
@@ -6,31 +6,44 @@ package services
...
 	return elasticSearchMasterService
 }
 
 // Creates Elasticsearch Data service element
 func createElasticsearchDataServiceElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) *corev1.Service {
-	elasticsearchDataService := createServiceElement(vmo, config.ElasticsearchData)
-
+	var elasticsearchDataService *corev1.Service = createServiceElement(vmo, config.ElasticsearchData)
 	// Data k8s service only needs to expose NodeExporter port
+	// - in dev mode, preserve this as the front end ES data port, at lest for now
```
bobdonat, line 40: typo lest

pkg/resources/helper.go
michael-cico: OK, yeah we weren't sure if we should align the flags or not.  Right now Doug and I are using the same flag, but they can be considered orthogonal features most likely.

pkg/resources/helper.go
michael-cico: I believe that's Doug's initials, it's basically a hack we're sharing until this is surfaced in the config.

pkg/resources/services/elasticsearch.go
```diff
@@ -6,31 +6,44 @@ package services
...
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // Creates Elasticsearch Client service element
 func createElasticsearchIngestServiceElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) *corev1.Service {
-	return createServiceElement(vmo, config.ElasticsearchIngest)
+	var elasticsearchIngestService *corev1.Service = createServiceElement(vmo, config.ElasticsearchIngest)
+	if resources.IsDevProfile(vmo) {
+		elasticsearchIngestService.Spec.Selector = resources.GetSpecID(vmo.Name, config.ElasticsearchMaster.Name)
+		elasticsearchIngestService.Spec.Ports = []corev1.ServicePort{resources.GetServicePort(config.ElasticsearchData)}
```
pfmackin, line 20: Is this using the same service port as the ES data?  If so won't you have a port conflict?

pkg/resources/statefulsets/statefulset.go
```diff
@@ -156,20 +178,38 @@ fi`,
...
+				Spec: corev1.PersistentVolumeClaimSpec{
+					AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{corev1.ResourceStorage: resource.MustParse(storageSize)},
+					},
+				},
+			}}
+	} else {
+		glog.Info("ES storage size" + vmo.Spec.Elasticsearch.Storage.Size)
+		//sizeLimit := resource.MustParse(vmo.Spec.Elasticsearch.Storage.Size)
```
bobdonat, line 203: can commented out code be removed

pkg/resources/services/elasticsearch.go
michael-cico: I believe they use separate front-end ports

vmi-system-es-data                        ClusterIP   10.96.32.87     <none>        9100/TCP            29h
vmi-system-es-ingest                      ClusterIP   10.96.112.95    <none>        9200/TCP            29h
vmi-system-es-master                      ClusterIP   10.96.254.208   <none>        9300/TCP            29h

pkg/resources/services/elasticsearch.go
```diff
@@ -6,31 +6,44 @@ package services
...
 // Creates Elasticsearch Data service element
 func createElasticsearchDataServiceElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) *corev1.Service {
-	elasticsearchDataService := createServiceElement(vmo, config.ElasticsearchData)
-
+	var elasticsearchDataService *corev1.Service = createServiceElement(vmo, config.ElasticsearchData)
 	// Data k8s service only needs to expose NodeExporter port
+	// - in dev mode, preserve this as the front end ES data port, at lest for now
 	elasticsearchDataService.Spec.Ports = []corev1.ServicePort{resources.GetServicePort(config.NodeExporter)}
+	if resources.IsDevProfile(vmo) {
+		// In dev mode, use the ES master pod, and target the main 9200 port
```
pfmackin, line 43: I don't understand this comment.  Aren't all the pods master/ingest/data?

pkg/resources/helper.go
michael-cico: I'll remove those.

pkg/resources/statefulsets/statefulset.go
```diff
@@ -156,20 +178,38 @@ fi`,
...
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{corev1.ResourceStorage: resource.MustParse(storageSize)},
+					},
+				},
+			}}
+	} else {
+		glog.Info("ES storage size" + vmo.Spec.Elasticsearch.Storage.Size)
```
bobdonat, line 202: need a space after size in log message

pkg/resources/services/elasticsearch.go
michael-cico: It's probably just a misnomer.  I'll fix the comment.

pkg/resources/statefulsets/statefulset.go
```diff
@@ -156,20 +178,38 @@ fi`,
...
+		//sizeLimit := resource.MustParse(vmo.Spec.Elasticsearch.Storage.Size)
+		statefulSet.Spec.Template.Spec.Volumes = []corev1.Volume{
+			{
+				Name: esMasterVolName,
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{ /*SizeLimit: &sizeLimit*/ },
```
bobdonat, line 208: same here. can comment be removed

pkg/resources/helper.go
michael-cico: I think for now, to avoid impacting Doug's and Tony's changes, we will keep it this for now to avoid any wiggle, we can look into renaming it next Sprint.  Plus, any env var will be replaced by config/CRD changes at some point soon when we design that.

pkg/resources/helper.go
michael-cico: Opened VZ-1680 to rename this.

markxnelson (approved):
those asserts in the statefulset_test feel like they could be improved - maybe with some kind of deep compare? but LGTM

PR #45, opened 2020-11-05
VZ-1315: zap structured logging

shihchang:
zap structured logging


go.mod
```diff
@@ -3,17 +3,17 @@ module github.com/verrazzano/verrazzano-monitoring-operator
 go 1.13
 
 require (
-	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/gorilla/mux v1.7.3
 	github.com/kylelemons/godebug v1.1.0
 	github.com/prometheus/client_golang v1.2.1
-	github.com/spf13/pflag v1.0.5
-	github.com/stretchr/testify v1.4.0
+	github.com/stretchr/testify v1.5.1
```
desagar, line 9: why did testify version change? I think it doesn't matter from an approval perspective since we only use it for tests, but curious why it changed

pkg/util/logs/logs.go
```diff
@@ -4,41 +4,31 @@
...
-}
-
-// NewLogger creates a new log.Logger which sends logs to glog.Info.
-func NewLogger(prefix string) *log.Logger {
-	return log.New(GlogWriter{}, prefix, 0)
+// InitLogs initializes logs with Time and Global Level of Logs set at Info
+func InitLogs(opts kzap.Options) {
+	var config zap.Config
+	if opts.Development {
+		config = zap.NewDevelopmentConfig()
```
desagar, line 16: We are calling InitLogs with empty zap.Options{} - what is the default value for opts.Development then? Also, how is this related to the devProfile stuff used by the ES code in statefulset.go?

go.mod
shihchang: I saw verrazzano-operator is using v1.5.1, and it is approved, so i updated it.

pkg/util/logs/logs.go
shihchang: opts.Development defaults to false. When opts.Development=true, it output to console instead of using json encoder

desagar: Also how do I set the option to something other than the default?

shihchang: verrazzano/install/chart/templates/03-verrazzano-monitoring-operator.yaml or edit deploy to add args ex.
```
args:
            - --zap-log-level=info
```
I will add this in my next PR on verrazzano/verrazzano

desagar: Ok thx - I would consider also adding the parameter for Development to false in the helm chart.

desagar (approved):

PR #49, opened 2020-11-18
VZ-1739.  Remove explicit dev profile hooks from verrazzano-monitoring-operator

michael-cico:
Remove explicit dev/prod profile checks, and let the verrazzano-operator drive the VMI configuration 
- We do still enforce single-node cluster behaviors when appropriate (only a single master node replica is configured)
- We also enforce that a valid multi-node cluster configuration consists of at least 1 of each type of node (1 master, 1 data, 1 ingest)
- The VO will drive single/multi-node ES and any persistent storage usage via the fine-grained VMI settings

### DIFF
--- a/eswait.go
+++ b/eswait.go
@@ -72,8 +72,10 @@ func (node ESNode) isSufficient(version string) bool {
 }
 
 func (node ESNode) isDataRole() bool {
-	// This should likely be a "contains" check as a node can be multiple roles
-	return strings.EqualFold(node.Role, "d")
+	// Check the role string for the data-node flag
+	// - role string consists of 1-3 chars, "d" (data), "i" (ingest), "m" (master), so look that it contains "d"
+	// - a node can be any combination of roles (e.g., single-node cluster would be "dim")
+	return strings.Contains(node.Role, "d")
 }
 
 // Send GET request to given URL, expected JSON response
--- a/eswait.go
+++ b/eswait.go
@@ -72,6 +72,7 @@ func (node ESNode) isSufficient(version string) bool {
 }
 
 func (node ESNode) isDataRole() bool {
+	// This should likely be a "contains" check as a node can be multiple roles
 	return strings.EqualFold(node.Role, "d")
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -23,10 +23,12 @@ func NewConfigFromConfigMap(configMap *corev1.ConfigMap) (*OperatorConfig, error
 	}
 	var config OperatorConfig
 	err := yaml.Unmarshal([]byte(configString), &config)
-	zap.S().Debugf("Unmarshalled configmap is:\n %s", configMap.String())
 	if err != nil {
 		return nil, fmt.Errorf("failed to unmarshall ConfigMap %s: %v", configMap.String(), err)
 	}
+	// Comment out the debug line below to avoid halting the debugger in Goland/IntelliJ
+	// see https://youtrack.jetbrains.com/issue/GO-8953
+	zap.S().Debugf("Unmarshalled configmap is:\n %s", configMap.String())
 
 	// Set defaults for any uninitialized values
 	zap.S().Infow("Setting config defaults")
--- a/pkg/resources/deployments/deployment.go
+++ b/pkg/resources/deployments/deployment.go
@@ -127,10 +127,12 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
 	}
 
 	// Elasticsearch
-	// - Hacking in the profile for now to skip the ingest/data deployment objects
-	// - We should likely use a factory pattern here to create different VMI profiles, although I think it might
-	//   be better to can the default specs as template files and read them in
-	if vmo.Spec.Elasticsearch.Enabled && !resources.IsDevProfile() {
+	// - V8O supports essentially 2 "known" configurations, a "prod" and a "dev" configuration for ES; while we want
+	//   to allow customizing topologies, we need to enforce certain constraints for now.
+	// - We are arbitrarily choosing to enforce that a "valid" multi-node cluster includes at least one separate
+	//   data node and one separate ingest node
+	// - This will weed out creating separate pods for data/ingest in the single-node cluster configuration as well
+	if vmo.Spec.Elasticsearch.Enabled && resources.IsValidMultiNodeESCluster(vmo) {
 		var es Elasticsearch = ElasticsearchBasic{}
 		deployments = append(deployments, es.createElasticsearchDeploymentElements(vmo, pvcToAdMap)...)
 	}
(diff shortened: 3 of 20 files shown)


eswait.go
```diff
@@ -72,6 +72,7 @@ func (node ESNode) isSufficient(version string) bool {
 }
 
 func (node ESNode) isDataRole() bool {
+	// This should likely be a "contains" check as a node can be multiple roles
```
pfmackin, line 75: Isn't this contains change needed for the laptop config?

pkg/config/config.go
```diff
@@ -23,10 +23,12 @@ func NewConfigFromConfigMap(configMap *corev1.ConfigMap) (*OperatorConfig, error
 	}
 	var config OperatorConfig
 	err := yaml.Unmarshal([]byte(configString), &config)
-	zap.S().Debugf("Unmarshalled configmap is:\n %s", configMap.String())
 	if err != nil {
 		return nil, fmt.Errorf("failed to unmarshall ConfigMap %s: %v", configMap.String(), err)
 	}
+	// Comment out the debug line below to avoid halting the debugger in Goland/IntelliJ
+	// see https://youtrack.jetbrains.com/issue/GO-8953
+	zap.S().Debugf("Unmarshalled configmap is:\n %s", configMap.String())
```
pfmackin, line 31: do you mean to delete the Debugf line  (based on your comment at line 29)

pkg/resources/deployments/deployment_test.go
```diff
@@ -90,13 +82,13 @@ func TestVMODevProfileFullDeploymentSize(t *testing.T) {
 			},
 			Elasticsearch: vmcontrollerv1.Elasticsearch{
 				Enabled:    true,
-				IngestNode: vmcontrollerv1.ElasticsearchNode{Replicas: 1},
+				IngestNode: vmcontrollerv1.ElasticsearchNode{Replicas: 0},
```
pfmackin, line 85: why is the replicas being set to 0 for all nodes.  Where are you setting the replica count to 1 for the single pod profile?

pkg/resources/deployments/deployment_test.go
```diff
@@ -106,6 +98,38 @@ func TestVMODevProfileFullDeploymentSize(t *testing.T) {
...
+			},
+			Elasticsearch: vmcontrollerv1.Elasticsearch{
+				Enabled:    true,
+				IngestNode: vmcontrollerv1.ElasticsearchNode{Replicas: 0},
+				MasterNode: vmcontrollerv1.ElasticsearchNode{Replicas: 0},
+				DataNode:   vmcontrollerv1.ElasticsearchNode{Replicas: 0},
+			},
+		},
+	}
+	assert.False(t, resources.IsSingleNodeESCluster(vmo), "Single node ES setup, expected IsDevProfile to be true")
```
pfmackin, line 127: shouldn't is assert true?   You are calling isSingleNodeESCCluster and expecting true correct?

eswait.go
michael-cico: Right now we don't include that container, I can see if I can get that fixed now.

pkg/config/config.go
michael-cico: No, the comment is really in case someone tries to run it in the debugger; this line can crash the run in the debugger, but not when run normally.

pkg/resources/deployments/deployment_test.go
michael-cico: For single-node cluster, we expect the following replicas:

Master: 1
Data: 0
Ingest: 0

for other configurations we expect at least one of each.

pkg/resources/deployments/deployment_test.go
michael-cico: No, the message in the assert is wrong...there's no replicas configured, so it's not a single-node cluster configuration.

michael-cico: Fixed the test

eswait.go
michael-cico: Fixed the eswait container, and updated the PR for the installer changes.

pkg/config/config.go
agv426: good to know, added that to the supportability wiki

pkg/resources/deployments/deployment.go
```diff
@@ -127,12 +128,18 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
...
+	// - V8O supports essentially 2 "known" configurations, a "prod" and a "dev" configuration for ES; while we want
+	//   to allow customizing topologies, we need to enforce certain constraints for now.
+	// - We are arbitrarily choosing to enforce that a "valid" multi-node cluster includes at least one separate
+	//   data node and one separate ingest node
+	// - This will weed out creating separate pods for data/ingest in the single-node cluster configuration as well
+	if vmo.Spec.Elasticsearch.Enabled {
+		if resources.IsValidMultiNodeESCluster(vmo) {
+			var es Elasticsearch = ElasticsearchBasic{}
+			deployments = append(deployments, es.createElasticsearchDeploymentElements(vmo, pvcToAdMap)...)
+		} else if !resources.IsSingleNodeESCluster(vmo) {
```
desagar, line 140: where do the deployments get created for single node ES cluster?

pkg/resources/deployments/deployment.go
```diff
@@ -156,18 +163,14 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
...
-				Name:  config.ESWait.Name,
-				Image: config.ESWait.Image,
-				// `-number-of-data-nodes 1` tells eswait to look for at least one data node
-				// `-timeout 5m` tells eswait to wait up to 5 minutes for desired state
-				Args: []string{"-number-of-data-nodes", "1", "-timeout", "5m", elasticsearchURL, config.ESWaitTargetVersion},
-			}
-			deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, waitForEsInitContainer)
-		} else {
-			zap.S().Infow("In development mode, bypassing ESWait container creation for Kibana")
+		waitForEsInitContainer := corev1.Container{
```
desagar, line 166: In the case of single node, wouldn't number of data nodes be 0?

michael-cico: The "master" node is really acting in all roles, master, data, ingest.  I had to modify the eswait container to be able to work correctly in a single-node cluster scenario.

pkg/resources/deployments/deployment.go
michael-cico: It's created in statefulset.go... the master is managed as a stateful set.

desagar (dismissed):

desagar (approved):

desagar (approved):

PR #52, opened 2020-12-21
fix scan issue

pfmackin:
Yum changes to fix security scan.  This same yum command is done for other Verrazzano images


docker-images/verrazzano-monitoring-instance-eswait/Dockerfile
```diff
@@ -27,6 +27,9 @@ RUN go build \
 
 FROM container-registry.oracle.com/os/oraclelinux:7-slim@sha256:fcc6f54bb01fc83319990bf5fa1b79f1dec93cbb87db3c5a8884a5a44148e7bb AS final
 
+RUN yum update -y && yum-config-manager --save --setopt=ol7_ociyum_config.skip_if_unavailable=true \
```
mgianatagh, line 30: Was there a newer oraclelinux:7-slim image that contains the security fix?

pfmackin: Probably, but all of the images have the same problem so I just used that fix.  Mark said we will do a pass to update all of the images at some point, but not for this pass.

gilbode (approved):

PR #54, opened 2021-01-07
VZ-1885.  Patch Elasticsearch

michael-cico:
Force ES/Prometheus to apply updates to work around rolling upgrade bug
- VMO attempts to do a rolling upgrade of Prometheus and ES nodes, however the ES data nodes are getting stuck on a diff in the RequestMemory that seems out of our control
- For now, allow the upgrade to continue at the same time, and get the rolling upgrade working later if necessary
- Update run-vmo.sh to pick up images from correct values.yaml location

### DIFF
--- a/pkg/vmo/deployment.go
+++ b/pkg/vmo/deployment.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2020, Oracle and/or its affiliates.
+// Copyright (C) 2020, 2021, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package vmo
@@ -7,7 +7,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	vmcontrollerv1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/config"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/constants"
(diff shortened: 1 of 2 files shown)


pkg/vmo/deployment.go
```diff
@@ -133,3 +133,20 @@ func updateNextDeployment(controller *Controller, vmo *vmcontrollerv1.Verrazzano
...
+
+// Temp hack, force all deployments to update regardless of diffs until we can figure out the rolling upgrade problem
+func updateAllDeployment(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, deployments []*appsv1.Deployment) (dirty bool, err error) {
+	for _, curDeployment := range deployments {
+		_, err := controller.deploymentLister.Deployments(vmo.Namespace).Get(curDeployment.Name)
+		if err != nil {
+			return false, err
+		}
+
+		// FIXME: on the next pass, we need to figure out why we are getting diffs in the RequestMemory that we don't expect
```
pfmackin, line 145: remove the Fixme

pkg/vmo/deployment.go
```diff
@@ -133,3 +133,20 @@ func updateNextDeployment(controller *Controller, vmo *vmcontrollerv1.Verrazzano
 	}
 	return false, nil
 }
+
+// Temp hack, force all deployments to update regardless of diffs until we can figure out the rolling upgrade problem
```
pfmackin, line 137: Remove the hack note, we will track internally.  Just comment that we are updating the deployments concurrently

pkg/vmo/deployment.go
michael-cico: fixed

pkg/vmo/deployment.go
michael-cico: fixed.

pfmackin (approved):

PR #59, opened 2021-02-11
VZ-2097.  Set ES logging to INFO by default.

michael-cico:
- Use CLI args to the ES containers to set the default logging level to INFO by default
- Also fix up the run-vmo.sh wrapper script

You can still enable other levels at runtime by posting new settings to ES, e.g.:

PUT /_cluster/settings
{
"persistent" :{ "logger.org.elasticsearch" : "DEBUG" }
}


bobdonat (approved):
LGTM

PR #63, opened 2021-03-04
oidc oauth reverse proxy sidecar

shihchang:
Add oidc oauth reverse proxy sidecar to elasticsearch, prometheus, kibana, and grafana
* redirect unauthenticated browser access to keycloak to login
* require Bearer type authorization header for API access
* continue to support Basic auth for API access

### DIFF
--- a/pkg/constants/oidc_proxy.go
+++ b/pkg/constants/oidc_proxy.go
@@ -1,0 +1,565 @@
+// Copyright (C) 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package constants
+
+// OidcCallbackPath is the callback URL path of OIDC authentication redirect
+const OidcCallbackPath = "/_authentication_callback"
+
+// OidcConfLua defines the conf.lua file name in OIDC proxy ConfigMap
+const OidcConfLua = "conf.lua"
+
+// OidcConfLuaTemp is the template of conf.lua file in OIDC proxy ConfigMap
+const OidcConfLuaTemp = `-- conf.lua
+    local ingressUri = 'https://'..'%s'
+    local oidcProviderHost = "%s"
+    local oidcProviderHostInCluster = "%s"
+    local realm ="%s"
+    local callbackPath = "%s"
+    local oidcClient = "%s"
+    local oidcDirectAccessClient = "%s"
+    local cookieKey = "%s"
+    local authStateTtlInSec = %v
+    local bearerServiceAccountToken = false
+
+    local oidcProviderInClusterUri = ""
+    if oidcProviderHostInCluster and oidcProviderHostInCluster ~= "" then
+        oidcProviderInClusterUri = 'http://'..oidcProviderHostInCluster..'/auth/realms/'..realm
+    end
+    local oidcProviderUri = 'https://'..oidcProviderHost.. '/auth/realms/'..realm
+    
+    local auth = require("auth").config({
+        ingressUri = ingressUri,
+        callbackPath = callbackPath,
+        callbackUri = ingressUri..callbackPath,
+        oidcProviderUri = oidcProviderUri,
+        oidcProviderAuthUri = oidcProviderUri..'/protocol/openid-connect/auth',
+        oidcProviderInClusterUri = oidcProviderInClusterUri,
+        oidcClient = oidcClient,
+        oidcDirectAccessClient = oidcDirectAccessClient,
+        authStateTtlInSec = authStateTtlInSec,
+        cookieKey = cookieKey,
+        bearerServiceAccountToken = bearerServiceAccountToken
+    })
+    ngx.header["Access-Control-Allow-Origin"] =  ngx.req.get_headers()["origin"]
+    ngx.header["Access-Control-Allow-Headers"] =  "authorization"
+    if ngx.req.get_method() == "OPTIONS" then
+        ngx.status = 200
+        ngx.exit(ngx.HTTP_OK)
+    end
+    auth.info("Extracting authorization header from request.")
+    local authHeader = ngx.req.get_headers()["authorization"]
+    if not authHeader then
+        if string.find(ngx.var.request_uri, callbackPath) then
+            auth.handleCallback()
+        end
+        if auth.authenticated() then
+            return
+        else
+            auth.authenticate()
+        end
+    else
+        auth.authHeader(authHeader) 
+    end
+`
+
+// OidcAuthLua defines the auth.lua file name in OIDC proxy ConfigMap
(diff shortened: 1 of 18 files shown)


shihchang:
The oidc proxy implementation is the lua scripts in [pkg/constants/oidc_proxy.go](https://github.com/verrazzano/verrazzano-monitoring-operator/pull/63/files#diff-f5b45aa5191c24022e072f8901214bae54a32904c90bc6b9a2b42a01abb09f45), embedded in a configMap to configure a nginx sidecar attached to each of the deployments.

pkg/constants/oidc_proxy.go
```diff
@@ -0,0 +1,540 @@
+// Copyright (C) 2020, 2021, Oracle and/or its affiliates.
```
pfmackin, line 1: remove 2020

pkg/constants/oidc_proxy.go
```diff
@@ -0,0 +1,540 @@
...
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package constants
+
+//Define the configuration templates for OIDC proxy
+const (
+	OidcCallbackPath = "/_authentication_callback"
```
pfmackin, line 8: formatting doesn't line up

pkg/constants/oidc_proxy.go
```diff
@@ -0,0 +1,540 @@
+// Copyright (C) 2020, 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package constants
+
+//Define the configuration templates for OIDC proxy
```
pfmackin, line 6: add space after //

pkg/constants/oidc_proxy.go
pfmackin, line 8: Where is this template coming from?   Can you put a doc link in the code?

pkg/resources/deployments/deployment.go
```diff
@@ -119,6 +122,11 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, operatorConfig *confi
 		deployment.Spec.Template.Spec.SecurityContext = &corev1.PodSecurityContext{
 			FSGroup: &grafanaGid,
 		}
+		if config.Grafana.OidcProxy != nil {
+			oidcConfig, oidcProxy := resources.CreateOidcProxy(vmo, &vmo.Spec.Grafana.Resources, &config.Grafana)
```
pfmackin, line 126: oidcConfig should be called oidcVolume to make it clear what it is.  Same comment throughout.

pkg/resources/ingresses/ingress.go
```diff
@@ -198,3 +214,55 @@ func noAuthOnHealthCheckSnippet(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance
 }
 `
 }
+
+//oidcProxyIngress creates the Ingress of the OidcProxy
+func oidcProxyIngress(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, component *config.ComponentDetails) *extensions_v1beta1.Ingress {
```
pfmackin, line 219: This should be newOidcProxyIngress or createOidcProxyIngress

pkg/resources/ingresses/ingress.go
```diff
@@ -119,26 +119,20 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) ([]*extensions_v1beta
 	ingresses = append(ingresses, ingress)
 
 	if vmo.Spec.Grafana.Enabled {
-		// Create Ingress Rule for Grafana Endpoint
-		ingRule = createIngressRuleElement(vmo, config.Grafana)
-		host := config.Grafana.Name + "." + vmo.Spec.URI
-		ingress, err := createIngressElementNoBasicAuth(vmo, host, config.Grafana, ingRule)
-		if err != nil {
-			return ingresses, err
+		if config.Grafana.OidcProxy != nil {
```
michael-cico, line 122: It mostly seems pretty straightforward, but what sets the OidcProxy configuration in the configmap for each component?  Is that going to be done by the `verrazzano-operator`?

shihchang: In pkg/config/components.go controlled by the OIDC_AUTH_ENABLED flage of env-var

pkg/constants/oidc_proxy.go
shihchang: removed

pkg/constants/oidc_proxy.go
shihchang: fixed

pkg/constants/oidc_proxy.go
shihchang: fixed

pkg/constants/oidc_proxy.go
shihchang: added godoc comment to each const items

pkg/resources/deployments/prometheus.go
```diff
@@ -186,6 +188,12 @@ func createPrometheusNodeDeploymentElements(vmo *vmcontrollerv1.VerrazzanoMonito
 		}
 		prometheusDeployment.Spec.Template.Spec.Volumes = append(prometheusDeployment.Spec.Template.Spec.Volumes, volume)
 
+		if config.Prometheus.OidcProxy != nil {
+			oidcConfig, oidcProxy := resources.CreateOidcProxy(vmo, &vmo.Spec.Prometheus.Resources, &config.Prometheus)
```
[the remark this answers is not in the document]
shihchang: fixed

pkg/resources/ingresses/ingress.go
shihchang: chagned to newOidcProxyIngress

michael-cico (approved):

pfmackin (approved):

wmhopkins (approved):
I'm approving this, but I have concerns about the approach to implementing the proxy.
- I'm not convinced that we've implented the OIDC protocols correctly/completely -- not saying we haven't, just that it needs a careful review to make sure, and I haven't had time to do that.
- I don't think we should be storing JWTs, or any part of them, in a cookie. If we really need to support sessions, we need to implement a session cookie. It may also be possible (and much better) to let Keycloak manage the sessions so we don't have to.
- With both this proxy and the api-proxy, I'm concerned that the code/configuration for nginx is potentially exposed because it's in a config map (including user accounts and passwords). Would be safer to build our own proxy image, I think, rather than rely on the generic one configured dynamically. We should also consolidate on a single proxy impl for both api-proxy and oidc-proxy.

Can you create a jira to look at these, or add them to the existing jira for authZ and rename it (to include more than authZ)?

PR #64, opened 2021-03-11
Add istio scrap targets to VMI system prometheus

bobdonat:
This pull request adds Istio scrap targets to the VMI system Prometheus.

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -220,6 +220,7 @@ func GetDefaultPrometheusConfiguration(vmo *vmcontrollerv1.VerrazzanoMonitoringI
 	dynamicScrapeAnnotation := prometheusValidLabelName + "_io_scrape"
 	namespace := vmo.Namespace
 	nginxNamespace := "ingress-nginx"
+	istioNamespace := "istio-system"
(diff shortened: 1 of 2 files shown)


pkg/resources/helper.go
```diff
@@ -279,7 +283,44 @@ scrape_configs:
      target_label: kubernetes_namespace
    - source_labels: [__meta_kubernetes_pod_name]
      action: replace
-     target_label: kubernetes_pod_name`)
+     target_label: kubernetes_pod_name
+
+ # Scrape config for Istio envoy stats
```
pfmackin, line 288: Curious if Istio aggregates envoy metrics for all the envoy sidecars and the aggregation is what is being scraped.

pfmackin (approved):

pkg/resources/helper.go
bobdonat: envoy metrics are per pod

pfmackin (changes requested):
Just noticed copyright needs to be changed

bobdonat:
> Just noticed copyright needs to be changed

Fixed.

pfmackin (approved):

PR #65, opened 2021-03-15
Qvdang/vz2300

qvdang:
Replacing with a more secure hash algorithm.

### DIFF
--- a/pkg/vmo/secret.go
+++ b/pkg/vmo/secret.go
@@ -1,12 +1,11 @@
-// Copyright (C) 2020, Oracle and/or its affiliates.
+// Copyright (C) 2020, 2021, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package vmo
 
 import (
 	"context"
-	"crypto/sha1"
-	"encoding/base64"
+	"golang.org/x/crypto/bcrypt"
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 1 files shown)


pkg/vmo/secret.go
```diff
@@ -50,9 +51,11 @@ func (hp HashedPasswords) Bytes() (passwordBytes []byte) {
 
 // SetPassword set a password for a user with a hashing algo
 func (hp HashedPasswords) SetPassword(name, password string) (err error) {
-	prefix := "{SHA}"
-	hash := hashSha(password)
-	hp[name] = prefix + hash
+	hashValue, ok := hashBcrypt(password)
+	if ok != nil {
```
pfmackin, line 55: can you use err, not ok

qvdang: Done.

pkg/vmo/secret.go
```diff
@@ -32,11 +31,13 @@ const (
 // HashedPasswords name => hash
 type HashedPasswords map[string]string
 
-func hashSha(password string) string {
-	s := sha1.New()
-	s.Write([]byte(password))
-	passwordSum := s.Sum(nil)
-	return base64.StdEncoding.EncodeToString(passwordSum)
+func hashBcrypt(password string) (string, error) {
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
```
wmhopkins, line 35: I assume `bcrypt.DefaultCost` is the number of rounds for the bcrypt algorithm? Did you research how many rounds is considered "secure", and compare that with the default value? What did you find?

wmhopkins (commented):
Looks good to me, but would like to understand the `DefaultCost` value relative to secure best practice. Do the GPS secure coding guidelines say anything about that?

pkg/vmo/secret.go
qvdang: Yes, DefaultCost is the cost factor(CF) currently set to 10 in the library.  It is the password expansion rounds(2^n).  Obviously, the higher the cost factor, the more secure the hash is.  However unless you know exactly what kind of hardware your env is running, it's not easy to choose a cost factor that balances out security and acceptable performance.  That was the reason I went with the default.  I did try various CF such as 14,15,16 but the responses were significantly slower on my OCI Linux host env.
GPS does not mention about the CF of the Bcrypt algorithm.  Let me know if you have any suggestion on the CF that we should use other than the default.

wmhopkins: Seems reasonable to me.

wmhopkins (approved):
Ship it!

PR #69, opened 2021-03-25
Delete the VMO from setting the default replicas to vmi-system-api

bhatpmk:
This change removes the code which sets the default replicas (which is 1) to vmi-system-api. This change is the first step to remove the dependency on verrazzano-monitoring-instance-api, further code clean-up will be handled separately.


pkg/vmo/vmospec.go
```diff
@@ -85,9 +85,6 @@ func InitializeVMOSpec(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMon
 	}
 
 	// Number of replicas for each component
-	if vmo.Spec.API.Replicas == 0 {
```
michael-cico, line 88: I wonder if this change alone is good enough?  If we allow the replica count to default to 0, then maybe we don't have to set it on the VO side.

michael-cico (approved):

pkg/vmo/vmospec.go
bhatpmk: @michael-cico I have not tried that, will test that out.

bhatpmk: @michael-cico I have confirmed that we don't need the change in Verrazzano Operator. Thank you.

pkg/vmo/vmospec.go
```diff
@@ -85,9 +85,6 @@ func InitializeVMOSpec(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMon
 	}
 
 	// Number of replicas for each component
```
pfmackin, line 87: you need to add 2021 to the copyright

bhatpmk: @pfmackin Thanks for noticing that, fixed now.

michael-cico (approved):

pfmackin (approved):

PR #70, opened 2021-03-31
Remove verrazzano-monitoring-instance-api image from Verrazzano 

tbeerbower:
Required for: VZ-2452

Changes to VMO to allow for the removal of the verrazzano-monitoring-instance-api image completely from Verrazzano `platform-operator/helm_config/charts/verrazzano/values.yaml`.

- Allow for optional components.
- Disable optional components if image is not specified.
- Make API component optional.
- Check for disabled flag for API component.
- Add unit test for optional components.

### DIFF
--- a/pkg/config/components.go
+++ b/pkg/config/components.go
@@ -5,6 +5,7 @@ package config
 
 import (
 	"fmt"
+	"go.uber.org/zap"
 	"os"
 	"strings"
 
@@ -26,6 +27,8 @@ type ComponentDetails struct {
 	RunAsUser         int64
 	EnvName           string
 	OidcProxy         *ComponentDetails
+	Optional          bool
+	Disabled          bool
 }
 
 // AllComponentDetails is array of all ComponentDetails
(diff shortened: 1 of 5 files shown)


pkg/config/components.go
```diff
@@ -216,7 +220,12 @@ func InitComponentDetails() error {
 		if len(component.EnvName) > 0 {
 			component.Image = os.Getenv(component.EnvName)
 			if len(component.Image) == 0 {
```
michael-cico, line 222: So, does `noimage` translate to an empty image string?  I know we set that in the helm chart, but didn't look to see how that was translated.

michael-cico (approved):

pkg/config/components.go
tbeerbower: Thanks for reviewing!

No, `noimage` doesn't translate to an empty string.  I thought about simply checking for "noimage" but the comments in the [PR for VZ-2255](https://github.com/verrazzano/verrazzano/pull/879) suggested that the `monitoringInstanceApiImage: "noimage"` line in the helm chart `values.yaml` should be removed altogether.  Also, the `alertManagerImage` component is specified as `noimage` so I didn't want to introduce a different meaning for it.

tbeerbower: In the follow up changes to Verrazzano I will remove the `monitoringInstanceApiImage: "noimage"` completely.

michael-cico: sounds good

PR #72, opened 2021-03-31
VZ-2347: Verrazzano Monitoring Instances (VMIs) should use Keycloak for authZ

fredtibbitts:
This adds role checking to the Lua sidecar scripts used in the VMIs. If the user does not have the "console_users" keycloak realm role, the user will see a 403 Forbidden error when attempting to access any of the VMI consoles.

I tested this with one user that does not have the role and one user that does have the role and verified the behavior is as expected. Once this PR is merged and a new VMO is built, I will submit a Verrazzano Platform Operator PR that pulls in the latest version of the VMO and also includes updates to the keycloak.json configuration to add the realm role and make it default.

### DIFF
--- a/pkg/constants/oidc_proxy.go
+++ b/pkg/constants/oidc_proxy.go
@@ -136,7 +136,7 @@ const OidcAuthLuaScripts = `-- auth.lua
       
     function me.unauthorized(msg, err)
         ngx.status = 401
-        me.logJson(ngx.ERR, meg, err)
+        me.logJson(ngx.ERR, msg, err)
         ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
       
@@ -210,7 +210,7 @@ const OidcAuthLuaScripts = `-- auth.lua
             local expiry = tonumber(ck.expiry)
             local refresh_expiry = tonumber(ck.refresh_expiry)
             if now > refresh_expiry then
-                -- refresh_token expired, redirt to authenticate
+                -- refresh_token expired, redirect to authenticate
                 me.authenticate()
             else
                 if now > expiry then
@@ -394,7 +394,7 @@ const OidcAuthLuaScripts = `-- auth.lua
         end
         local publicKey = me.publicKey(jwt_obj.header.kid)
         if not publicKey then
-            me.unauthorized("No public_key retreived from keycloak")
+            me.unauthorized("No public_key retrieved from keycloak")
         end
         local verified = jwt:verify_jwt_obj(publicKey, jwt_obj)
         if (tostring(jwt_obj.valid) == "false" or tostring(jwt_obj.verified) == "false") then
@@ -402,7 +402,7 @@ const OidcAuthLuaScripts = `-- auth.lua
         end
         me.logJson(ngx.INFO, "Check for groups in jwt token.")
         if ( not (jwt_obj.payload) or not (jwt_obj.payload.groups)) then
-            me.unauthorized("No groups asscoiated with user.")
+            me.unauthorized("No groups associated with user.")
         end
         ngx.req.clear_header("Authorization")
         if jwt_obj.payload and jwt_obj.payload.email then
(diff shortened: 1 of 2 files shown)


pkg/constants/oidc_proxy.go
```diff
@@ -265,7 +292,19 @@ const OidcAuthLuaScripts = `-- auth.lua
...
         end
     end
-      
+
+    function me.logout()
+        local redirectArgs = ngx.encode_args({
+            redirect_uri = me.logoutCallbackUri
+        })
+        local redirURL = me.oidcProviderUri.."/protocol/openid-connect/logout?"..redirectArgs
+        ngx.redirect(redirURL)
```
shihchang, line 301: Do it need to also cleanup the cookies here?

pkg/constants/oidc_proxy.go
```diff
@@ -135,11 +155,18 @@ const OidcAuthLuaScripts = `-- auth.lua
...
-        ngx.status = 401
-        me.logJson(ngx.ERR, meg, err)
+        ngx.status = ngx.HTTP_UNAUTHORIZED
+        me.logJson(ngx.ERR, msg, err)
         ngx.exit(ngx.HTTP_UNAUTHORIZED)
     end
       
+    function me.forbidden(msg, err)
+        ngx.status = ngx.HTTP_FORBIDDEN
+        me.deleteCookie("authn")
```
wmhopkins, line 165: We should probably be deleting the cookie for UNAUTHORIZED (which ironically really means unauthenticated), because that indicates we couldn't authenticate you. With FORBIDDEN, we're not letting you see a particular resource, but that doesn't mean you aren't legitimately authenticated, and removing your cookie effectively logs you out (from the locally managed session, not the Keycloak session, which is another issue with this code).

pkg/vmo/configmap.go
```diff
@@ -298,10 +298,12 @@ func oidcConfLuaScripts(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, compon
 		oidcProviderHostInCluster,
 		"verrazzano-system", //realm
 		constants.OidcCallbackPath,
+		constants.OidcLogoutCallbackPath,
```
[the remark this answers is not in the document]
fredtibbitts: I agree that templating should probably be used here, but the `Sprintf` was already in place and I don't want to change this all over to use templating as part of this PR.

pkg/constants/oidc_proxy.go
fredtibbitts: The logout callback function is deleting the authn cookie. I'm not sure if it makes much of a difference if it's deleted in the callback or before the redirect?

wmhopkins (commented):
This mostly looks good. I see some issues with this code as a whole, but apart from a couple small comments your changes look good.

I don't see where the identity is propagated back to Grafana, though -- unless I'm missing something, this is adding logout support, but not Grafana integration.

pkg/constants/oidc_proxy.go
fredtibbitts: The problem with not deleting the cookie in this case is that the user is now stuck. If the user again tries to access the VMI UI they will just get a 403 again, they don't have the opportunity to log in as a different user, and worse, if the role is subsequently added in keycloak for the user, it won't matter because the presence of the authn cookie means we won't go back to keycloak. It seems like a better user experience if we delete the cookie and force another login.

I did add code that logs the user out of keycloak, which removes the keycloak session token. This way if the role is added to the user after a 403, the next time they load the VMI UI they can re-authenticate and the identity token will have the role.

fredtibbitts:
@wmhopkins this PR does not include the Grafana identity propagation. I'm working on that now and I'll submit that in a separate PR.

pkg/constants/oidc_proxy.go
wmhopkins: The presence of the authn cookie is problematic, we should get rid of it, but that's a different issue.

I agree with you about the user experience. I guess we can leave it as is for now. Normally you wouldn't want to, because there might be other resources that the user _does_ have access to (e.g., the vz console itself). But there's a bunch of cleanup that needs to happen here, we can do fix this later.

wmhopkins (approved):

PR #73, opened 2021-04-01
VZ-2347: Add support for asserting user identity to Grafana

fredtibbitts:
This PR extends the VMI auth proxy to assert the user's identity to Grafana. The changes include overriding Grafana config to enable auth, setting an nginx variable based on the OIDC username, and adding a proxy header with that username. Grafana then uses that username and creates a local user.

I tested this with the verrazzano user and also created a new user in keycloak and it looks like it's doing the right thing. Grafana shows the current user in the lower left of the window, and you can view the user profile.


wmhopkins (approved):
LGTM.

PR #75, opened 2021-04-06
VZ-2347: More authZ updates

fredtibbitts:
This PR adds role checking and Grafana identity propagation to the basic auth path. Previously I was only doing the role check and identity propagation for the VMI console UI flow. After looking through the code and talking to Will, I added the same role check and identity propagation to the basic auth flow. This fixes hitting the Grafana API using basic auth.

I refactored the Lua script a bit to support code re-use in the multiple auth paths. I also removed the roles and username from the "authn" cookie.


wmhopkins (approved):
lgtm, ship it.

PR #77, opened 2021-04-08
VMO OIDC sidecar re-route requests to admin keycloak in managed cluster

guoyongzhang:
when the cluster registration secret is present, ie, the cluster is a managed cluster, the oidc sidecar would set oidcProviderHost with admin keycloak host; the ca-bundle is from registration secret and mounted to sidecar.

### DIFF
--- a/pkg/resources/deployments/deployment.go
+++ b/pkg/resources/deployments/deployment.go
@@ -20,12 +20,12 @@ import (
 
 // Elasticsearch interface
 type Elasticsearch interface {
-	createElasticsearchDeploymentElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, pvcToAdMap map[string]string, isManaged bool) []*appsv1.Deployment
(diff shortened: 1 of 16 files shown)


pkg/resources/deployments/deployment.go
```diff
@@ -20,12 +20,12 @@ import (
 
 // Elasticsearch interface
 type Elasticsearch interface {
-	createElasticsearchDeploymentElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, pvcToAdMap map[string]string) []*appsv1.Deployment
+	createElasticsearchDeploymentElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, pvcToAdMap map[string]string, isManaged bool) []*appsv1.Deployment
```
shihchang, line 23: Should the "isManaged" be added to VerrazzanoMonitoringInstance/VerrazzanoMonitoringInstanceSpec as a property?

guoyongzhang: updated.

shihchang (approved):

PR #78, opened 2021-04-13
proxy-body-size

shihchang:
Set proxy-body-size which is default to 1m and caused "413 request size exceed" for fluentd

### DIFF
--- a/pkg/resources/ingresses/ingress.go
+++ b/pkg/resources/ingresses/ingress.go
@@ -186,7 +186,9 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) ([]*extensions_v1beta
 	}
 	if vmo.Spec.Elasticsearch.Enabled {
 		if config.ElasticsearchIngest.OidcProxy != nil {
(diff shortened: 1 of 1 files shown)


pkg/resources/ingresses/ingress.go
```diff
@@ -186,7 +186,9 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) ([]*extensions_v1beta
 	}
 	if vmo.Spec.Elasticsearch.Enabled {
 		if config.ElasticsearchIngest.OidcProxy != nil {
-			ingresses = append(ingresses, newOidcProxyIngress(vmo, &config.ElasticsearchIngest))
+			ingress := newOidcProxyIngress(vmo, &config.ElasticsearchIngest)
+			ingress.Annotations["nginx.ingress.kubernetes.io/proxy-body-size"] = "65M"
```
guoyongzhang, line 190: why this one is 65M and the other one is 6M?  maybe should document the reason in comments.  should 65M be defined in const as well?

guoyongzhang (approved):

PR #80, opened 2021-04-16
Changes related to node-exporter

bobdonat:
This pull request adds a missing Prometheus scrape target for node-exporter metrics and removes the node-exporter containers from Prometheus and the Elasticsearch data pod.  The node-exporter containers are not needed since we deploy node-exporter as a daemonset.


pkg/resources/services/elasticsearch.go
```diff
@@ -37,9 +37,6 @@ func createElasticsearchMasterServiceElements(vmo *vmcontrollerv1.VerrazzanoMoni
 // Creates Elasticsearch Data service element
 func createElasticsearchDataServiceElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) *corev1.Service {
 	var elasticsearchDataService *corev1.Service = createServiceElement(vmo, config.ElasticsearchData)
-	// Data k8s service only needs to expose NodeExporter port
-	// - in dev mode, preserve this as the front end ES data port, at least for now
-	elasticsearchDataService.Spec.Ports = []corev1.ServicePort{resources.GetServicePort(config.NodeExporter)}
```
pfmackin, line 42: I assume this means we can now remove 9100 ingress from ES network policy

bobdonat: Yes, that will be in the next PR for the verrazzano repo to pull in an updated montoring-operator image

pfmackin (approved):

PR #84, opened 2021-05-06
Remove the deployment of the Prometheus Gateway

bobdonat:
This pull request removes the deployment of the Prometheus Gateway since it is no longer needed.
In addition, the VMO CRD was synced up with what is being installed with the Verrazzano helm chart.

Addresses VZ-2081


k8s/crds/verrazzano.io_verrazzanomonitoringinstances_crd.yaml
```diff
@@ -270,6 +249,46 @@ spec:
             uri:
               type: string
               pattern: '[A-Za-z0-9-_]+'
+            externalSiteMonitor:
```
jmaron99, line 252: Not sure what to make of this addition?

pkg/resources/helper.go
```diff
@@ -367,12 +359,6 @@ func New64Val(value int64) *int64 {
 	return &val
 }
 
-// NewBool return a pointer to a boolean given a boolean value
```
jmaron99, line 370: looks like some general cleanup here, correct?

THIRD_PARTY_LICENSES.txt
```diff
@@ -1434,4 +1434,4 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 ATTRIBUTION-HELPER-GENERATED:
-License file based on go.mod with md5 sum: 8133ec1a24f927eaddbe2b436650f583
+License file based on go.mod with md5 sum: 7b3c256b55a3ebbc24c2b988e7881ee1
```
desagar, line 1437: Weird that TPL didn't produce additional entries for the dependency you added on oci-go-sdk. Do you know why that is?

k8s/crds/verrazzano.io_verrazzanomonitoringinstances_crd.yaml
desagar: ditto :)

bobdonat: I do not know what this addition is all about.  I made the crd yaml equivalent to what we are installing with the verrazzano helm chart,

pkg/resources/helper.go
bobdonat: yes, this function was not used.

k8s/crds/verrazzano.io_verrazzanomonitoringinstances_crd.yaml
jmaron99: so the monitoring operator crd that gets installed as part of the installation includes this?  something that isn't included in the actual VMO crd?

THIRD_PARTY_LICENSES.txt
bobdonat: The thirdpartylicense check had failed on my initial push in the pipeline with the following so I ran the attribution-helper 

Third party license file ends with 'License file based on go.mod with md5 sum: 8133ec1a24f927eaddbe2b436650f583' expected 'License file based on go.mod with md5 sum: 7b3c256b55a3ebbc24c2b988e7881ee1'

This was the change

k8s/crds/verrazzano.io_verrazzanomonitoringinstances_crd.yaml
bobdonat: and this the crd yaml that we are not generating

bobdonat: I thought it would be better to make the CRD yamls the same instead of have two variants

THIRD_PARTY_LICENSES.txt
desagar: hm, maybe it doesn't include it if the copyright owner is Oracle?

desagar (approved):

PR #90, opened 2021-05-12
Remove code that serves VMO metrics for Prometheus

bobdonat:
Addresses Jira VZ-2455.

This pull request removes the code that was needed to get two VMO prometheus metrics which we don't scrape for.


pkg/vmo/controller.go
```diff
@@ -412,15 +405,6 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
 		c.clusterInfo.KeycloakCABundle = clusterSecret.Data[constants.KeycloakCABundleData]
 	}
 
-	// If lock, controller will not sync/process the VMO env
-	labels := prometheus.Labels{"namespace": vmo.Namespace, "vmo_name": vmo.Name}
-	if vmo.Spec.Lock {
```
pfmackin, line 417: Is lock only related to metrics?

bobdonat: This is the code that references Lock.  I am not clear what it is used for.   I do not see the value being set anywhere in the VMO code.

			// If lock, controller will not sync/process the VMO env
	labels := prometheus.Labels{"namespace": vmo.Namespace, "vmo_name": vmo.Name}
	if vmo.Spec.Lock {
		zap.S().Infof("[%s/%s] Lock is set to true, this VMO env will not be synced/processed.", vmo.Name, vmo.Namespace)
		metrics.Lock.With(labels).Set(1)
		return nil
	}
	metrics.Lock.Delete(labels)

bobdonat: Looks like it.  The field is not referenced anywhere in VMO.

bobdonat: Do you think I should remove Lock from the CRD?

bobdonat: fixed.  I had removed too much code.

pfmackin (approved):

PR #93, opened 2021-05-26
Scrape NGINX ingress controller metrics only on the port specified by prometheus.io/port

bhatpmk:
This PR changes renames the job kubernetes-pods to nginx-ingress-controller, after removing the verrazzano-system namespace from the job. As part of this change, the job configuration is updated to scrape NGINX ingress controller metrics only on port 10254. As part of NGINX ingress controller installation, Verrazzano overrides the chart value prometheus.io/port to set "10254".

### DIFF
--- a/pkg/resources/helper_test.go
+++ b/pkg/resources/helper_test.go
@@ -43,12 +43,12 @@ func TestGetDefaultPrometheusConfiguration(t *testing.T) {
 	role = kubernetesSdConfigs.([]interface{})[0].(map[interface{}]interface{})["role"]
 	assert.Equal(t, "pod", role, "kubernetes_sd_configs should have - role: pod")
 
-	kubernetesPods := getItem("job_name", "kubernetes-pods", scrapeConfigs.([]interface{}))
-	assert.NotNil(t, kubernetesPods)
-	kubernetesSdConfigs = kubernetesPods["kubernetes_sd_configs"]
+	ingressController := getItem("job_name", "nginx-ingress-controller", scrapeConfigs.([]interface{}))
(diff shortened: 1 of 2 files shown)


pkg/resources/helper_test.go
```diff
@@ -42,6 +42,15 @@ func TestGetDefaultPrometheusConfiguration(t *testing.T) {
 	kubernetesSdConfigs = envoyStats["kubernetes_sd_configs"]
 	role = kubernetesSdConfigs.([]interface{})[0].(map[interface{}]interface{})["role"]
 	assert.Equal(t, "pod", role, "kubernetes_sd_configs should have - role: pod")
+
+	kubernetesPods := getItem("job_name", "kubernetes-pods", scrapeConfigs.([]interface{}))
```
bobdonat, line 46: I think the job name should be changed to nginx-ingress-controller instead of kubernetes-pods to reflect the metrics being scraped.

bobdonat:
I think there is a need to get a nginx metrics during the ATs to verify nginx metrics are being scraped.  If we had this before this bug have been caught during an AT run.

bhatpmk:
> I think there is a need to get a nginx metrics during the ATs to verify nginx metrics are being scraped. If we had this before this bug have been caught during an AT run.

@bobdonat I am adding a test case to query sample nginx metrics like "nginx_ingress_controller_success" from Prometheus.

pkg/resources/helper_test.go
bhatpmk: I have renamed the job.

jmaron99 (approved):
Reviewed over zoom

PR #94, opened 2021-06-15
VZ-2688: Have oauth-proxy use shared proxy code

wmhopkins:
Use shared proxy code to generate configmaps for VMI oauth-proxy instances. As an expedient, check in a copy of the shared proxy code from the verrazzano repo; this will be removed and replaced with a dependency on the verrazzano repo proxy package in a separate PR.


pkg/proxy/proxy_config.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (C) 2020, 2021, Oracle and/or its affiliates.
+// Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+package proxy
```
shihchang, line 4: Should not this reuse https://github.com/verrazzano/verrazzano/tree/master/pkg/proxy?

pkg/vmo/configmap.go
```diff
@@ -8,10 +8,10 @@ import (
 	"context"
 	"fmt"
 	"html/template"
-	"math/rand"
 	"net/url"
 	"strings"
-	"time"
+
+	proxy "github.com/verrazzano/verrazzano-monitoring-operator/pkg/proxy"
```
shihchang, line 14: Should not this proxy import/reuse "github.com/verrazzano/verrazzano/pkg/proxy"?

pkg/proxy/auth_lua_template.go
```diff
@@ -0,0 +1,784 @@
...
+
+    local logoutCallbackUri = nil
+
+    function me.config(opts)
+        for key, val in pairs(opts) do
+            me[key] = val
+        end
+        if not cookiekey then
+            -- this is a global variable, initialize it exactly once
+            cookiekey = me.randomBase64(32)
```
shihchang, line 49: If the proxy restarted for any reason, it will use a new cookiekey making all the cookies previously created on the client side invalid. Does it handle this condition?

markxnelson (approved):

PR #99, opened 2021-06-30
change default JVM min max for ES

pfmackin:
The JVM args for ES nodes (master, ingest, data) were hardcoded and there was no way to change them.  I modified the code to derive them from the pod.request for each of the 3 ES node types.  The JVM value is calculated to be 75% of the pod request size to account for overhead.  The JVM min/max must be identical or else ES pods won't start.  

Also, I adjusted the default master from 512m to 700m to be conservative.

### DIFF
--- a/pkg/util/memory/jvmheap.go
+++ b/pkg/util/memory/jvmheap.go
@@ -22,7 +22,7 @@ func FormatJvmHeapMinMax(heap string) string {
 	return fmt.Sprintf("-Xms%s -Xmx%s", heap, heap)
 }
 
-// Format the string based on the size of the input value
+// FormatJvmHeapSize formats the string based on the size of the input value
 // Return whole number (1200M not 1.2G)
 // Return min magnatude of K
 func FormatJvmHeapSize(sizeB int64) string {
--- a/pkg/util/memory/jvmheap.go
+++ b/pkg/util/memory/jvmheap.go
@@ -17,18 +17,18 @@ const UnitM = 1024 * UnitK
 const UnitG = 1024 * UnitM
 
 // FormatJvmHeapMinMax returns the identical min and max JVM heap setting in the format
-// java expects like "-Xms2g -Xmx2g"
+// Java expects like "-Xms2g -Xmx2g"
 func FormatJvmHeapMinMax(heap string) string {
 	return fmt.Sprintf("-Xms%s -Xmx%s", heap, heap)
 }
 
 // FormatJvmHeapSize formats the string based on the size of the input value
 // Return whole number (1200M not 1.2G)
-// Return min magnatude of K
+// Return a minimum magnitude of K
 func FormatJvmHeapSize(sizeB int64) string {
 	if sizeB >= UnitG {
 		if sizeB%UnitG == 0 {
-			// e.g 50g
+			// e.g. 50g
 			return fmt.Sprintf("%.0fg", float64(sizeB)/UnitG)
 		}
 		if sizeB%UnitM == 0 {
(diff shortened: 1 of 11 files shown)


pkg/util/memory/jvmheap.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if sizeB%UnitK == 0 {
+		// e.g. 1500k - value in exact kb
+		return fmt.Sprintf("%.0fk", float64(sizeB)/UnitK)
+	}
+	// round up to next kb
+	return fmt.Sprintf("%vk", sizeB/UnitK+1)
+}
+
+//func FormatJvmHeapSize(sizeB int64) string {
```
michael-cico, line 53: Cruft?

michael-cico (dismissed):

pkg/util/memory/jvmheap.go
pfmackin: fixed

michael-cico (approved):

pkg/util/memory/jvmheap.go
```diff
@@ -0,0 +1,51 @@
...
+const UnitK = 1024
+
+// UnitM is 1 Megabyte
+const UnitM = 1024 * UnitK
+
+// UnitG is 1 Gigabyte
+const UnitG = 1024 * UnitM
+
+// FormatJvmHeapMinMax returns the identical min and max JVM heap setting in the format
+// java expects like "-Xms2g -Xmx2g"
```
markxnelson, line 20: Java has a capital J

pkg/util/memory/jvmheap.go
```diff
@@ -0,0 +1,51 @@
...
+func FormatJvmHeapMinMax(heap string) string {
+	return fmt.Sprintf("-Xms%s -Xmx%s", heap, heap)
+}
+
+// FormatJvmHeapSize formats the string based on the size of the input value
+// Return whole number (1200M not 1.2G)
+// Return min magnatude of K
```
markxnelson, line 27: spelling

pkg/util/memory/jvmheap.go
```diff
@@ -0,0 +1,51 @@
...
+func FormatJvmHeapSize(sizeB int64) string {
+	if sizeB >= UnitG {
+		if sizeB%UnitG == 0 {
+			// e.g 50g
```
markxnelson, line 31: e.g.

pkg/util/memory/jvmheap.go
```diff
@@ -0,0 +1,51 @@
...
+			return fmt.Sprintf("%.0fg", float64(sizeB)/UnitG)
+		}
+		if sizeB%UnitM == 0 {
+			// e.g. 1500m - value in exact mb
```
markxnelson, line 35: mb, kb, etc., are captialized

markxnelson (commented):
couple of small issues with spelling/caps/punctuation - code is fine

markxnelson (approved):

pkg/util/memory/jvmheap.go
pfmackin: either one works for JVM heap args, VMO was using lower case so i used that.

pkg/util/memory/jvmheap.go
pfmackin: fixed

pkg/util/memory/jvmheap.go
pfmackin: fixed

pkg/util/memory/jvmheap.go
pfmackin: fixed

michael-cico (approved):

PR #130, opened 2021-11-11
VZ-3747 fix - delete the Prometheus config map at startup rather than overwrite

desagar:
Overwriting the Prometheus config map is causing it to change constantly since other operators write to it as well. Instead, delete the config map at VMO startup so that it gets re-created once by VMO and the edits from other operators stay instead of disappearing.


cmd/verrazzano-monitoring-ctrl/main.go
```diff
@@ -60,6 +60,9 @@ func main() {
 
 	vmo.StartHTTPServer(controller)
 
+	// Run any cleanup that needs to happen pre-startup of the controller
```
mgianatagh, line 63: copyright date needs to be updated

mgianatagh (approved):

PR #148, opened 2022-02-25
Recreate Kibana on upgrade

anders-swanson:


bobdonat:
I think a description of the change would be useful

bobdonat (approved):

michael-cico (approved):

PR #149, opened 2022-02-25
Recreate Kibana on upgrade

anders-swanson:
We recreate kibana on upgrade, since kibana does not support a rolling upgrade. If multiple kibana versions are running you may have a data race and subsequent data corruption.

### DIFF
--- a/pkg/resources/deployments/deployment.go
+++ b/pkg/resources/deployments/deployment.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2020, Oracle and/or its affiliates.
+// Copyright (C) 2020, 2022, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package deployments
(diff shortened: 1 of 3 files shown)


gilbode:
This seems like it will result in downtime anytime the kibana deployment is rolled out, not just for kibana version upgrades.  Any idea how much downtime & what other circumstances besides upgrade of verrazzano this would happen?

Also - this is fixing a bug.  We somehow need a test for this.

anders-swanson:
> This seems like it will result in downtime anytime the kibana deployment is rolled out, not just for kibana version upgrades. Any idea how much downtime & what other circumstances besides upgrade of verrazzano this would happen?
> 
> Also - this is fixing a bug. We somehow need a test for this.

- Yes, it will result in downtime. About 30-60 seconds for Kibana. We could potentially mitigate this for changes that do not require a recreate by dynamically selecting the Strategy: Only use recreate if the Kibana image is changing. 
- Testing that Recreate is set as the strategy during an upgrade is probably sufficient

gilbode:
> * Testing that Recreate is set as the strategy during an upgrade is probably sufficient

Can you please add a unit test to this PR that tests that?

anders-swanson:
Added tests and set to use recreate only if Kibana image is changing.

gilbode (approved):

bobdonat:
same comment as other PR - I think a description of the change would be useful

michael-cico (approved):

PR #150, opened 2022-03-04
VZ 5212: Enable monitoring code-gen

anders-swanson:
code-gen target was outdated. Updated code-gen and types.


pkg/apis/vmcontroller/v1/types.go
```diff
@@ -94,7 +94,8 @@ type (
 		Resources              Resources `json:"resources,omitempty"`
 		RetentionPeriod        int32     `json:"retentionPeriod,omitempty"`
 		Replicas               int32     `json:"replicas,omitempty"`
-		HTTP2Enabled           bool      `json:"http2Enabled" yaml:"http2Enabled"`
+		// +optional
+		HTTP2Enabled bool `json:"http2Enabled,omitempty"`
```
pfmackin, line 98: looks like formatting out of alignment

pfmackin: I see you got rid of YAML, not sure that was intentional

anders-swanson: will reset this, it is fine as is

Makefile
```diff
@@ -49,19 +49,37 @@ INTEG_RUN_REGEX=Test
...
 BUILDVERSION=`git describe --tags`
 BUILDDATE=`date +%FT%T%z`
 
 .PHONY: code-gen
-code-gen:
-	git config core.hooksPath hooks
-	GO111MODULE=on $(GO) mod vendor
-	chmod +x vendor/k8s.io/code-generator/*.sh
-	sh hack/update-codegen.sh
+code-gen: controller-gen
```
pfmackin, line 63: we use generate and manifest targets for VPO.  Can you do the same?

Makefile
```diff
@@ -49,19 +49,37 @@ INTEG_RUN_REGEX=Test
 INGRESS_CONTROLLER_SVC_NAME:=ingress-controller
 GO ?= go
 HELM_CHART_NAME ?= verrazzano-monitoring-operator
-
+CONTROLLER_GEN_VERSION:=v0.4.1
```
mgianatagh, line 52: The makefile for the platform-operator uses the following approach to get the version.  This approach may require few edits when the version changes.  Just edit the go.mod.

CONTROLLER_GEN_VERSION ?= $(shell go list -m -f '{{.Version}}' sigs.k8s.io/controller-tools)

mgianatagh (dismissed):

pfmackin (approved):

PR #153, opened 2022-03-08
VZ 4923: Allow comms on port 9200 to master node

anders-swanson:


theabmitra (approved):
Approving it.  As a reminder ensure to add the network policy check as well as part of a separate PR.

michael-cico (approved):

PR #156, opened 2022-03-10
VZ 4923/5251: Start dashboards after OpenSearch is Ready.

anders-swanson:
- Start Kibana/OSD after OpenSearch is ready
- Add Per-Node Storage configuration and StorageClass for OpenSearch/ES nodes


pkg/resources/statefulsets/statefulset_test.go
```diff
@@ -20,13 +20,15 @@ import (
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
 )
 
+const defaultStorageClass = "default"
```
gilbode, line 23: Is there some significance in Kubernetes to the name "default"?  i.e. will this always use the default storage class?

anders-swanson: This variable is used for testing only.

pkg/apis/vmcontroller/v1/types.go
```diff
@@ -122,6 +125,7 @@ type (
 		Replicas  int32     `json:"replicas,omitempty"`
 		JavaOpts  string    `json:"javaOpts" yaml:"javaOpts,omitempty"`
 		Resources Resources `json:"resources,omitempty"`
+		Storage   *Storage  `json:"storage,omitempty"`
```
michael-cico, line 128: So this is largely to expose the size right?  We're not planning to expose the AD and PVC names from VZ?

anders-swanson: Yes. I am planning to use AD/PVC in a later task (VMO configurability) where nodes may have more than one role

anders-swanson:
The companion PR to this fix (in VPO), needs to set the OpenSearch version to 1.2.3 in the VMO helm chart template.

pkg/resources/deployments/deployment.go
```diff
@@ -156,10 +156,38 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, kubeclientset kuberne
...
+		deployment.Spec.Template.Spec.Containers[0].Env = []corev1.EnvVar{
+			{Name: "VMI_NAME", Value: vmo.Name},
+			{Name: "NAMESPACE", Value: vmo.Namespace},
+			{Name: "ENV_NAME", Value: operatorConfig.EnvName},
+		}
+		if len(vmo.Spec.NatGatewayIPs) > 0 {
+			deployment.Spec.Template.Spec.Containers[0].Args = []string{fmt.Sprintf("--natGatewayIPs=%s", strings.Join(vmo.Spec.NatGatewayIPs, ","))}
+		}
+
+		deployment.Spec.Template.Spec.Containers[0].LivenessProbe.InitialDelaySeconds = 15
```
pfmackin, line 174: where did you get these values?

anders-swanson: May make more sense if you view the files side by side or use split view

pkg/vmo/pvc.go
```diff
@@ -37,14 +37,14 @@ func createPersistentVolumeClaims(controller *Controller, vmo *vmcontrollerv1.Ve
 		controller.log.Errorf("Failed to create PVC specs for VMI %s: %v", vmo.Name, err)
 		return nil, err
 	}
-	deploymentToAdMap := map[string]string{}
+	pvcToAdMap := map[string]string{}
```
pfmackin, line 40: what is an AD?  Can update the function comment and spell it out

pkg/vmo/deployment.go
```diff
@@ -19,6 +19,39 @@ import (
...
 
+func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) error {
+	if osd == nil {
+		return nil
+	}
+
+	var err error
+	existingDeployment, err := controller.deploymentLister.Deployments(vmo.Namespace).Get(osd.Name)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
pfmackin, line 30: Don't you have to wait for OpenSearch to be ready in the case where this is a new OSD?

pkg/vmo/pvc.go
anders-swanson: Availability domain, corresponding to the underlying node. This was existing before my change, I renamed the var to pvcToAdMap since it was not a deployment to Ad Map.

pkg/vmo/deployment.go
anders-swanson: We could, gilbode had suggested only during upgrade. OSD _should_ be able to start without waiting for OS if it is a clean install.

gilbode: When we used to wait at install time it greatly increased install time.  Personally I think it's unnecessary.

pkg/vmo/pvc.go
pfmackin: We shouldn't need to worry about AD if we have anti affinity set.  Anyway, you answered my question

pkg/resources/pvcs/pvc.go
```diff
@@ -14,6 +14,7 @@ import (
 // New will return a new Service for VMO that needs to executed for on Complete
 func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, storageClassName string) ([]*corev1.PersistentVolumeClaim, error) {
 	var pvcList []*corev1.PersistentVolumeClaim
+	dataNodeStorage := resources.GetStorageForNode(vmo, vmo.Spec.Elasticsearch.DataNode)
```
pfmackin, line 17: this can be nil, can't it?  Storage is a pointer in the CRD spec

anders-swanson: see
```
//GetStorageForNode selects the Storage object to use, prioritizing the node-level configuration
// Spec.Elasticsearch.Storage exists for historical reasons, and we use this function to ensure compatibility
func GetStorageForNode(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, node vmcontrollerv1.ElasticsearchNode) *vmcontrollerv1.Storage {
	if node.Storage != nil {
		return node.Storage
	}
	return &vmo.Spec.Elasticsearch.Storage
}
```

Elasticsearch.Storage is not a pointer.

anders-swanson: This func is keep API compatibility. Eventually we can version the API and get rid of Elasticsearch.Storage

pfmackin: I was looking at
Storage   *Storage  `json:"storage,omitempty"`

anders-swanson: Right. That's the one I added on the per-node level

pfmackin (approved):

PR #161, opened 2022-03-15
PVC Resize

anders-swanson:
Implements PVC resize algorithm for OpenSearch when the underlying storage class does not allow volume expansion.

### DIFF
--- a/pkg/vmo/health.go
+++ b/pkg/vmo/health.go
@@ -38,32 +38,49 @@ var doHTTP = func(client *http.Client, request *http.Request) (*http.Response, e
 	return client.Do(request)
 }
 
+func IsOpenSearchUpgradeable(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) error {
+	return opensearchHealth(vmo, true)
+}
+
 //IsOpenSearchReady verifies the of the OpenSearch Cluster is ready to use by checking the cluster status is green,
 // and that each node is running the expected version
-func IsOpenSearchReady(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) (bool, error) {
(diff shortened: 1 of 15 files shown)


pkg/vmo/health.go
```diff
@@ -30,14 +30,22 @@ type (
...
+	MinDataNodesForResize = 2
 )
 
 var doHTTP = func(client *http.Client, request *http.Request) (*http.Response, error) {
 	return client.Do(request)
 }
 
+func IsOpenSearchResizable(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) error {
+	if vmo.Spec.Elasticsearch.DataNode.Replicas < MinDataNodesForResize {
+		return fmt.Errorf("cannot resize OpenSearch with less than 3 data nodes. Scale up your cluster 3 or more data nodes")
```
pfmackin, line 44: shouldn't error say < 2 nodes

pkg/vmo/pvc.go
```diff
@@ -66,35 +66,47 @@ func createPersistentVolumeClaims(controller *Controller, vmo *vmcontrollerv1.Ve
 		controller.log.Debugf("Applying PVC '%s' in namespace '%s' for VMI '%s'\n", pvcName, vmo.Namespace, vmo.Name)
 		existingPvc, err := controller.pvcLister.PersistentVolumeClaims(vmo.Namespace).Get(pvcName)
 
-		// If the PVC already exists, we *only* read its current AD, *if possible* (this is not possible for all storage classes and situations)
+		// If the PVC already exists, we check if it needs resizing
 		if existingPvc != nil {
```
pfmackin, line 70: better to continue if existingPvc == nil so you can get rid of an indentation level

anders-swanson: There's an else block for when the existingPVC is nil

PR #164, opened 2022-03-17
VZ 5327: ISM Plugin for Verrazzano

anders-swanson:
VMO to configure the ISM plugin for Verrazzano, if ISM is enabled.


k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
```diff
@@ -146,6 +148,15 @@ spec:
                     type: object
                   enabled:
                     type: boolean
+                  indexManagement:
+                    properties:
+                      enabled:
+                        type: boolean
+                      maxApplicationIndexAge:
```
ayaranga, line 155: We plan to add configuration support for other index management operations such as rollover in future. Also, it would be good to group such configurations and add it under either application or system or any other type we want to support in future. My suggestion would be something like:
```
indexManagement:
  system:
    actions:
      delete:
        minAge: 7d
      rollover:
        minAge: 1d
        minDocCount: 10000
        minSize: 10G
  application:
    actions:
      delete:
        minAge: 3d
      rollover:
        minAge: 1d
...
```

pkg/vmo/health.go
```diff
@@ -39,6 +36,12 @@ var doHTTP = func(client *http.Client, request *http.Request) (*http.Response, e
 	return client.Do(request)
 }
 
+func resetDoHTTP() {
```
kminder, line 39: Is this intended purely for testing.  In any case please document why this would be called.

k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
```diff
@@ -63,6 +63,7 @@ spec:
                         pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                         type: string
                       maxSizeDisk:
+                        description: These fields are not used anywhere
```
kminder, line 66: Shouldn't there also be a max storage limit?  What happens if the volumes fill bill before rollover deletion occurs?

k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
```diff
@@ -146,6 +148,15 @@ spec:
                     type: object
                   enabled:
                     type: boolean
+                  indexManagement:
```
kminder, line 151: This API change should be reviewed/discussed.
Is this already exposed in the VZ CRD?

k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
```diff
@@ -146,6 +148,15 @@ spec:
...
+                    properties:
+                      enabled:
+                        type: boolean
+                      maxApplicationIndexAge:
+                        type: string
+                      maxSystemIndexAge:
+                        type: string
```
kminder, line 158: The string encoded durations should have details about how units are handled.

k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
anders-swanson: Not exposed, and yes, this does need review.

k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
anders-swanson: The operator doesn't use these fields at all. I had added the description to match

k8s/crds/verrazzano.io_verrazzanomonitoringinstances.yaml
anders-swanson: See updated configuration

pkg/resources/ism/ism.go
```diff
@@ -0,0 +1,319 @@
...
+	if !policyNeedsUpdate(policy, existingPolicy) {
+		return nil, nil
+	}
+
+	payload, err := serializeIndexManagementPolicy(policy)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Println(string(payload))
```
kminder, line 145: Is this debug?

pkg/resources/ism/ism_test.go
```diff
@@ -0,0 +1,298 @@
...
+		MinIndexAge:  &age,
+		Rollover: vmcontrollerv1.RolloverPolicy{
+			MinIndexAge: &rolloverAge,
+			MinSize:     &minSize,
+			MinDocCount: &minDocCount,
+		},
+	}
+}
+
+func makeISMVMI(age string, enabled bool) *vmcontrollerv1.VerrazzanoMonitoringInstance {
```
kminder, line 88: Minor nit about consistency.  create vs make vs new

pkg/resources/ism/ism_test.go
```diff
@@ -0,0 +1,298 @@
...
+					*createTestPolicy(age, age, "*", "1gb", 1),
+				},
+			},
+		},
+	}
+
+	return v
+}
+
+func TestConfigureIndexManagementPluginISMDisabled(t *testing.T) {
```
kminder, line 106: The name of this is pretty descriptive but an you add the normal test doc for consistency.

pkg/resources/ism/ism.go
anders-swanson: Yep, thanks

pkg/resources/ism/ism_test.go
```diff
@@ -0,0 +1,298 @@
...
+	assert.NoError(t, <-Configure(&vmcontrollerv1.VerrazzanoMonitoringInstance{}))
+}
+
+func TestConfigureIndexManagementPluginHappyPath(t *testing.T) {
```
kminder, line 110: What is this actually testing?

pkg/resources/ism/ism_test.go
```diff
@@ -0,0 +1,298 @@
...
+			return &http.Response{
+				StatusCode: http.StatusOK,
+				Body:       io.NopCloser(strings.NewReader("")),
+			}, nil
+		}
+	}
+	vmi := makeISMVMI("1d", true)
+	ch := Configure(vmi)
+	assert.NoError(t, <-ch)
+	resetDoHTTP()
```
kminder, line 133: Are these resetDoHTTP calls going to interfere with each other when these test methods are run in parallel?

anders-swanson: Since we are using this to mock the HTTP client calls it wouldn't play nice with parallelism. The alternative to that is to pass the HTTP function in, although then you are passing around this function all the time.

anders-swanson: I could update it to be a bit smarter and parallel safe if you think it is worthwhile.

pkg/resources/ism/ism.go
```diff
@@ -0,0 +1,318 @@
...
+	"encoding/json"
+	"fmt"
+	vmcontrollerv1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
+	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
+	"net/http"
+	"reflect"
+	"strings"
+)
+
+var doHTTP = func(client *http.Client, request *http.Request) (*http.Response, error) {
```
ayaranga, line 17: Can we move this to common code so that it can be reused in other places as well?

kminder (approved):
Approved.

PR #165, opened 2022-03-18
Fix deprecated warnings for CustomResourceDefinitions and CronJobs

bobdonat:
This pull request removes the following warnings from repeatedly being logged.

W0318 12:48:16.538738       1 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition

W0317 22:19:13.507845       1 warnings.go:67] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob

CustomResourceDefinition was updated to v1 and CronJobs code was removed since it was not needed.


mgianatagh (approved):

pkg/vmo/controller.go
```diff
@@ -59,8 +58,6 @@ type Controller struct {
 	clusterRolesSynced   cache.InformerSynced
 	configMapLister      corelistersv1.ConfigMapLister
 	configMapsSynced     cache.InformerSynced
-	cronJobLister        batchlistersv1beta1.CronJobLister
```
pfmackin, line 62: why did this cronJob support get removed?

bobdonat: The code in cronjobs.go was only used by unit tests so I removed it.  The references in controller.go caused deprecation warnings.   Do we need VMO to handle cronJobs?

pfmackin (approved):

PR #166, opened 2022-03-22
VZ-5377: Opensearch backup preparedness

theabmitra:
This is required for backing up OpenSearch. This will allow the repository-s3 plugin to backup indicess to an S3 compatible object store


pkg/constants/constants.go
```diff
@@ -227,3 +227,12 @@ const KeycloakCABundleData = "ca-bundle"
 // PrometheusClusterNameLabel - the label name attached to all metrics to indicate the Verrazzano
 // cluster name where the metric originated
 const PrometheusClusterNameLabel = "verrazzano_cluster"
+
+const (
+	// Constants required for updating Opensearch keystore
```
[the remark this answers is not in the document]
theabmitra: This will backup to any S3 compatible storage  like minio./Aws S3 and OCI object store.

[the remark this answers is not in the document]
theabmitra: let me rename the variables

pkg/resources/statefulsets/statefulset.go
```diff
@@ -64,6 +83,32 @@ func createElasticsearchMasterStatefulSet(log vzlog.VerrazzanoLogger, vmo *vmcon
 		// HTTP is enabled on the master here solely for our readiness check below (on _cluster/health)
 		{Name: "HTTP_ENABLE", Value: "true"},
 		{Name: "logger.org.opensearch", Value: "info"},
+		{Name: constants.OciAccessKeyVarName,
```
michael-cico, line 86: Would it be better if these were mounted as optional volumes?

pkg/constants/constants.go
```diff
@@ -227,3 +227,12 @@ const KeycloakCABundleData = "ca-bundle"
...
+	VerrazzanoSecretName = "verrazzano"
+	OciAccessKeyVarName  = "OCI_ACCESS_KEY_ID"
```
michael-cico, line 234: had a similar comment in the VMO PR, but should the naming be indicative of S3 since we're using that compatibility layer?   It is OCI under the covers for now, but it should work with any S3-compatible storage right?

pkg/constants/constants.go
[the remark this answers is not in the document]
theabmitra: updated

pkg/constants/constants.go
theabmitra: yes its renamed

pkg/resources/deployments/elasticsearch.go
```diff
@@ -191,8 +191,54 @@ func (es ElasticsearchBasic) createElasticsearchDataDeploymentElements(vmo *vmco
...
+			"sh",
+			"-c",
+			`#!/usr/bin/env bash -e
+
+# Updating elastic search keystore with keys
+# required for the repository-s3 plugin
+
+if [ "${OBJECT_STORE_ACCESS_KEY_ID:-}" ]; then
+    echo "Updating oci access key..."
+	echo $OCI_ACCESS_KEY_ID | /usr/share/opensearch/bin/opensearch-keystore add --stdin --force s3.client.default.access_key;
```
pfmackin, line 233: should this be echo $OBJECT_STORE_ACCESS_KEY_ID

pkg/resources/deployments/elasticsearch.go
```diff
@@ -191,8 +191,54 @@ func (es ElasticsearchBasic) createElasticsearchDataDeploymentElements(vmo *vmco
...
+fi
+if [ "${OBJECT_STORE_ACCESS_SECRET_KEY_ID:-}" ]; then
+    echo "Updating oci secret access key..."
+	echo OCI_SECRET_ACCESS_KEY_ID | /usr/share/opensearch/bin/opensearch-keystore add --stdin --force s3.client.default.secret_key;
```
pfmackin, line 237: should this be echo $OBJECT_STORE_ACCESS_SECRET_KEY_ID

theabmitra: Thanks looks like a type when i updated :)

pkg/resources/statefulsets/statefulset.go
theabmitra: So the script does a check whether this value is availble. With this being enforced atleast we will see a pod configuration error. Otherwise things mayfail silently and even the script wont complain

pfmackin (dismissed):

pfmackin (approved):

PR #169, opened 2022-03-23
VZ-5396: Renaming secrets to be more backup specific

theabmitra:
Renaming secrets to be more backup specific


pkg/constants/constants.go
```diff
@@ -230,9 +230,9 @@ const PrometheusClusterNameLabel = "verrazzano_cluster"
 
 const (
 	// Constants required for updating Opensearch keystore
-	VerrazzanoSecretName          = "verrazzano"
+	VerrazzanoKubeObjectName      = "verrazzano-backup"
```
kminder, line 233: This seems a poor variable name.  Does there need to be two variables?

theabmitra: We cannot have the word "Secret" in a variable lint. thats flaged by golint

pfmackin: Can you use xyzScrt or something.

kminder: I was thinking it needed to have Backup in it somewhere.  This is a fairly broadly shared constants file right?

theabmitra: yes, updated the secret name to have satisfy both the criteria. safe name and backup

kminder (approved):
Approved.

PR #170, opened 2022-03-25
VZ 5328: Reindex to data streams on upgrade

anders-swanson:
Supports reindexing indices to data streams on upgrade. Reindexing happens IFF:
- Target data streams are present on the OpenSearch cluster (meaning the cluster is ready to reindex)
- Source indices are present on the OpenSearch cluster (meaning the cluster hasn't been fully reindexed yet)

After reindex, any index patterns matching old indices are updated to match the corresponding data streams.


pkg/opensearch/opensearch_upgrade.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	Dest struct {
+		Index  string `json:"index"`
+		OpType string `json:"op_type"`
+	}
+)
+
+// Reindex old style indices to data streams and delete it
+func (o *OSClient) MigrateIndicesToDataStreams(log vzlog.VerrazzanoLogger, vmi *vmcontrollerv1.VerrazzanoMonitoringInstance, openSearchEndpoint string) error {
```
pfmackin, line 68: if o is immutable then all the methods should use (o OSClient)

anders-swanson: All the functions that use properties of o are methods.

pkg/resources/helper.go
```diff
@@ -510,3 +520,24 @@ func OidcProxyService(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, componen
 		},
 	}
 }
+
+// convertToRegexp converts index pattern to a regular expression pattern.
+func ConvertToRegexp(pattern string) string {
```
kminder, line 525: Don't see a unit test for this.

pkg/constants/constants.go
```diff
@@ -98,12 +98,15 @@ const MetricsNameSpace = "vmo_operator"
 // DefaultPrometheusRetentionPeriod default Prometheus retention configuration
 const DefaultPrometheusRetentionPeriod = 90
 
-// ESHttpPort default Elasticsearch HTTP port
-const ESHttpPort = 9200
+// ESHTTPPort default Elasticsearch HTTP port
```
michael-cico, line 101: Just a nit, one one of these passes we should start replacing the ES name for OS... looks like you've done it in some cases.

pkg/opensearch/opensearch_upgrade.go
```diff
@@ -0,0 +1,327 @@
...
+	log.Debugf("OpenSearch: Found Verrazzano system indices %v", systemIndices)
+	return systemIndices
+}
+
+func getApplicationIndices(log vzlog.VerrazzanoLogger, indices []string) []string {
+	var appIndices []string
+	for _, index := range indices {
+		systemIndex := false
+		if strings.HasPrefix(index, "verrazzano-namespace-") {
+			for _, systemNamespace := range config.SystemNamespaces() {
```
michael-cico, line 146: Generally, this is going to be a static set right?  You could probably compute/initialize a package-private var and do a map lookup of the name instead of iterating, but I'm guessing this code isn't called that often anyway.  Just a thought...

pkg/constants/constants.go
michael-cico: Good times, am I right?

pkg/opensearch/opensearch_upgrade.go
anders-swanson: Yeah I wish this 1970s-style language had a full featured collections library.

pkg/resources/helper.go
anders-swanson: Added one. Thanks

pkg/opensearch/opensearch_upgrade.go
```diff
@@ -0,0 +1,323 @@
...
+var (
+	secondsPerMinute = uint64(60)
+	secondsPerHour   = secondsPerMinute * 60
+	secondsPerDay    = secondsPerHour * 24
+	secondsPerWeek   = secondsPerDay * 7
+	unitMultipliers  = map[uint8]uint64{
+		's': 1,
+		'm': secondsPerMinute,
+		'h': secondsPerHour,
+		'H': secondsPerHour,
```
kminder, line 30: Why a capitol only for hours?

kminder:
What makes the calls to reindexAndDeleteIndices or updatePatternsInternal idempotent? If VMO crashes during either of these will the next VMO reconcile cause data loss and/or inconsistencies?

kminder (approved):
Approved.

michael-cico (dismissed):

anders-swanson:
Added monitor to background the reindex operator - should be ready for re-review

michael-cico (approved):

PR #171, opened 2022-03-29
Disable deleting the prometheus configmap on vmo startup

lake-of-dreams:
Prevents customizations done to prometheus configmap before vmo restart to be lost after any reboot.


kminder:
This seems like something that there should be a test for.

lake-of-dreams:
Test added in https://github.com/verrazzano/verrazzano/pull/2839

pkg/vmo/configmap.go
```diff
@@ -15,6 +16,7 @@ import (
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/constants"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources/configmaps"
+	"gopkg.in/yaml.v2"
```
kminder, line 19: Do we have a PLS approval to use this?

pkg/vmo/configmap_test.go
```diff
@@ -50,32 +49,6 @@ func TestCreateConfigmaps(t *testing.T) {
 	assert.Equal(t, 8, len(all.Items))
 }
 
-// TestDeleteConfigMapIfExists tests the deleteConfigMapIfExists function
```
kminder, line 53: This change deletes a test but doesn't add any tests.  There needs to be a unit test for reconcilePrometheusConfigMap

lake-of-dreams: Added tests

pkg/vmo/configmap.go
lake-of-dreams: This [already](https://github.com/verrazzano/verrazzano-monitoring-operator/blob/0dfe30610f3c6b765fd1a990ab392751df72da46/pkg/config/config.go#L10) being used in this repo and is in TPL so I assume yes!

kminder (approved):
Approved.

PR #180, opened 2022-04-04
VZ-5360 create Grafana admin user from secret

mbenmclean:
Passes data from the grafana-admin secret to Grafana to create the verrazzano user with the admin role


wmhopkins (commented):
This looks good as far as passing in the secret instead of username/password, but I don't see where the secret is created.

mbenmclean:
It's created in the platform-operator. It will be part of my merge to the verrazzano repo

wmhopkins (approved):
lgtm.

PR #185, opened 2022-04-08
Vz 5659 - Labelling for PVC created from deployments

theabmitra:


pkg/resources/pvcs/pvc.go
```diff
@@ -46,11 +48,21 @@ func createPvcElements(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, vmoStor
...
 		return nil, err
 	}
+	resourceLabel := resources.GetMetaLabels(vmo)
+	if vmo.Spec.Prometheus.Enabled {
+		resourceLabel[constants.ComponentLabel] = resources.GetCompLabel(config.Prometheus.Name)
+	}
+	if vmo.Spec.Elasticsearch.Enabled {
+		resourceLabel[constants.ComponentLabel] = resources.GetCompLabel(config.ElasticsearchData.Name)
+	}
+	if vmo.Spec.Grafana.Enabled {
```
anders-swanson, line 58: Isn't this going to add multiple labels for each component? like if prometheus and grafana are both enabled, then all PVC will have prometheus + grafana labels? Seems like the PVC should only be labelled by the owning component.

theabmitra: fixed

anders-swanson (approved):
LGTM can you please try upgrade before merging

anders-swanson (approved):

PR #186, opened 2022-04-11
VZ 5649: Support multiple roles for a node group

anders-swanson:
Multiple roles are supported via selectors. The service selectors are updated to use `role-<rolename: true` as the selector.
The selectors are only updated after every pod in the opensearch cluster has been updated with its new labels - this avoids switching the selector and removing pods from the load balancer (taking down the cluster)


pkg/vmo/deployment.go
```diff
@@ -126,10 +127,14 @@ func CreateDeployments(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMon
 	}
 	for _, deployment := range existingDeploymentsList {
 		if !contains(deploymentNames, deployment.Name) {
-			controller.log.Debugf("Deleting deployment %s", deployment.Name)
-			err := controller.kubeclientset.AppsV1().Deployments(vmo.Namespace).Delete(context.TODO(), deployment.Name, metav1.DeleteOptions{})
-			if err != nil {
-				controller.log.Errorf("Failed to delete deployment %s: %v", deployment.Name, err)
+			if deployments.IsOpenSearchDataDeployment(vmo.Name, deployment) && expected.OpenSearchDataDeployments > 0 {
```
michael-cico, line 130: So, if it's an OS data node, and the expected data deployments are > 0, we allow the scale down.  Does this allow us to scale down to 0 deployments?  and do we need to prevent that?

anders-swanson: We explicitly need to allow it, since how else would you delete/clean up your cluster?

anders-swanson: there is an edge case though... If they have master+data nodes (STS) and deployment data nodes, and scale down their deployments to 0 -> all are deleted, and cluster potentially goes red.

michael-cico (approved):

michael-cico (approved):

PR #190, opened 2022-04-18
Bounce single node on update

anders-swanson:
To scale up a single node cluster, the node must be bounced to pick up the new cluster properties. This is necessary because clusters set to single node discovery cannot add nodes.

### DIFF
--- a/pkg/resources/statefulsets/plan.go
+++ b/pkg/resources/statefulsets/plan.go
@@ -45,7 +45,7 @@ func CreatePlan(log vzlog.VerrazzanoLogger, existingList, expectedList []*appsv1
 		ExistingCluster: mapping.existingSize > 0,
 		// Bounce the master node if it's a single node cluster and we're scaling up to a healthy size.
 		// This removes of the single node property in the pod environment.
-		BounceNodes: mapping.existingSize == 1 && mapping.expectedSize > 2,
+		BounceNodes: mapping.existingSize == 1 && mapping.expectedSize >= minClusterSize,
 	}
 	for name, expected := range mapping.expected {
 		existing, ok := mapping.existing[name]
--- a/pkg/resources/statefulsets/plan.go
+++ b/pkg/resources/statefulsets/plan.go
@@ -24,6 +24,7 @@ type (
 		Delete          []*appsv1.StatefulSet
 		Conflict        error
 		ExistingCluster bool
+		BounceNodes     bool
 	}
 
 	statefulSetMapping struct {
@@ -31,6 +32,7 @@ type (
 		expected           map[string]*appsv1.StatefulSet
 		isScaleDownAllowed bool
 		existingSize       int32
+		expectedSize       int32
 	}
 )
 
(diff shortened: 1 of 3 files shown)


pkg/resources/statefulsets/plan.go
```diff
@@ -41,6 +43,9 @@ func CreatePlan(log vzlog.VerrazzanoLogger, existingList, expectedList []*appsv1
 	plan := &StatefulSetPlan{
 		// There is no running cluster if the existing size is 0
 		ExistingCluster: mapping.existingSize > 0,
+		// Bounce the master node if it's a single node cluster and we're scaling up to a healthy size.
+		// This removes of the single node property in the pod environment.
+		BounceNodes: mapping.existingSize == 1 && mapping.expectedSize > 2,
```
kminder, line 48: Should this be mapping.expectedSize > 1 or >= 2?

anders-swanson: I put it as greater than 2 since a steady state cluster with only 2 master nodes is not a valid configuration - leader election cannot be performed with 2 master nodes.

anders-swanson: This should probably ready >= minClusterSize

pkg/vmo/statefulset.go
```diff
@@ -91,17 +91,23 @@ func getInitialMasterNodes(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, exi
...
 	}
 
-	_, err := c.kubeclientset.AppsV1().StatefulSets(vmo.Namespace).Update(context.TODO(), sts, metav1.UpdateOptions{})
-	return err
+	if _, err := c.kubeclientset.AppsV1().StatefulSets(vmo.Namespace).Update(context.TODO(), sts, metav1.UpdateOptions{}); err != nil {
+		return err
+	}
+	// if it was a single node cluster, delete the pod to ensure it picks up the updated settings.
+	if bounceNodes {
+		return c.kubeclientset.CoreV1().Pods(vmo.Namespace).Delete(context.TODO(), sts.Name+"-0", metav1.DeleteOptions{})
```
michael-cico, line 108: how is the update defined?  Wondering if it's better to execute the equivalent of a rollout restart, by setting an annotation like the kubectl command does.  Then the old pod will be taken down after the new one is spun up, possibly?  Unless it doesn't matter for this case.

anders-swanson: this is specifically for the single node case - the STS is already trying to upgrade, but it wont go through since the existing pod is healthy and it wont remove all of its ready replicas.
this is a tactical nuke to force the single node master to restart and remove it's single node discovery property.

anders-swanson: maybe there's a better way to do it. Do you think the annotation would make the ready replica restart?

michael-cico: We have other code to replicate a rollout restart by setting an annotation.  I think it depends on the deployment/sts policy.

It's a nit, probably doesn't matter for the single node case.

anders-swanson: I didn't go with annotations because:
- scale up adds the new nodes STS is 1/3. The new nodes will NEVER get to ready status because the 1 ready node is set to single node discovery
- that ready node will never be turned off by the STS controller until the others are ready, due to its update strategy.

kminder (approved):
Approved.

PR #192, opened 2022-04-20
VZ-5719: Change a couple of Grafana settings when Auth Proxy is disabled

fredtibbitts:
When Auth Proxy is disabled, users will access the Grafana web UI directly using HTTP basic auth. This PR disables allowing user sign up (the admin user can create users) and changes the default org role from Admin to Editor. The second change is probably not necessary since the admin will assign the role when creating users, but better to downgrade the role from a security standpoint.


wmhopkins (approved):
lgtm.

stevshap (approved):

PR #194, opened 2022-04-27
VZ-5645 Added prometheus job to scrape opensearch metrics

anajosep:
Signed-off-by: anajosep <anand.francis.joseph.augustin@oracle.com>

Changes for adding prometheus scrape config for scraping Opensearch metrics exported by the opensearch-prometheus-exporter plugin

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -483,16 +483,23 @@ scrape_configs:
      insecure_skip_verify: true
    metrics_path: "/_prometheus/metrics"
    kubernetes_sd_configs:
-   - role: service
+   - role: pod
      namespaces:
        names:
          - "` + constants.VerrazzanoSystemNamespace + `"
    relabel_configs:
-   - source_labels: [__meta_kubernetes_service_name]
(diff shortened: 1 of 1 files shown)


pkg/resources/helper.go
```diff
@@ -471,6 +471,38 @@ scrape_configs:
    - source_labels: null
      action: replace
      target_label: ` + constants.PrometheusClusterNameLabel + `
+     replacement: ` + vzClusterName + `"
+
+ # Scrape config for opensearch
+ - job_name: 'opensearch'
+   scheme: https
+   tls_config:
```
anders-swanson, line 479: just curious do we support scrape if istio is disabled

anders-swanson (approved):

pkg/resources/helper.go
anajosep: If Istio, is disabled, this scraping config would not work and would fail. In cases when Istio is disabled, `http` scheme without `tls_config` section needs to be used. Should I add that conditionally? I thought Istio is enabled always for VMI instances

ayaranga (approved):

PR #196, opened 2022-05-10
Reorder conditional to prevent array out of bounds

anders-swanson:
Array out of bounds could happen in some circumstances. Reordering the conditional prevents this.


kminder (approved):
Approved.

PR #198, opened 2022-05-11
Scale in Statefulset/Deployment

anders-swanson:
If the Statefulset or Deployment is down (no ready replicas), allow the resource to be updated or deleted.
This allows the use case where the user may want to scale-in or update a resource, but the cluster is not in a healthy state.
If the nodes are running, a safe scale down will be issued, to re-allocate any shards on the node.

### DIFF
--- a/pkg/vmo/deployment.go
+++ b/pkg/vmo/deployment.go
@@ -127,7 +127,7 @@ func CreateDeployments(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMon
 	}
 	for _, deployment := range existingDeploymentsList {
 		if !contains(deploymentNames, deployment.Name) {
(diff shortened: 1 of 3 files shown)


pkg/vmo/deployment.go
```diff
@@ -127,7 +127,7 @@ func CreateDeployments(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMon
 	}
 	for _, deployment := range existingDeploymentsList {
 		if !contains(deploymentNames, deployment.Name) {
-			if deployments.IsOpenSearchDataDeployment(vmo.Name, deployment) && expected.OpenSearchDataDeployments > 0 {
+			if deployments.IsOpenSearchDataDeployment(vmo.Name, deployment) && (expected.OpenSearchDataDeployments > 0 || deployment.Status.ReadyReplicas > 0) {
```
kminder, line 130: Some additional comments explaining the multilayer if logic here would be helpful.

kminder (approved):
Approved.

michael-cico (approved):

PR #202, opened 2022-05-18
VZ 5927: Scale up single node dev profile

anders-swanson:
In addition to allowing InstallArg profiles to scale up, this prevents a few nasty scenarios that lose cluster state:
- removing master roles dumps cluster state
- removing data roles causes shard allocation issues

There are ways to repurpose nodes, but the use of a CLI tool is required, run from within the node. It's recommended to create new nodes instead, and decommission old nodes with the undesired roles.

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -30,9 +30,9 @@ var (
 
 const serviceClusterLocal = ".svc.cluster.local"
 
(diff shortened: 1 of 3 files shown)


anajosep (approved):

kminder (approved):
Approved.

PR #207, opened 2022-05-31
VZ-6016: Remove Prometheus deployment, configmaps, etc.

fredtibbitts:
Remove the VMO-managed Prometheus as it is being replaced with the Prometheus Operator-managed Prometheus. I also removed most of the Alertmanager code since Alertmanager is going to be managed by Prometheus Operator too.

Note there are a few lines of Prometheus and Alertmanager left, dealing with ingress. I don't yet know if/where the ingress code is moving to, so I left it for now.

DO NOT merge until all of the other work to migrate off of the VMO Prometheus is complete.


pkg/constants/constants.go
```diff
@@ -137,48 +134,9 @@ const DashboardConfig = "dashboards"
...
-const PrometheusConfig = "prometheus-config"
-
-// PrometheusConfigVersions Prometheus config versions
-const PrometheusConfigVersions = "prometheus-config-versions"
-
-// PrometheusConfigMountPath Prometheus config mountpath
-const PrometheusConfigMountPath = "/etc/prometheus/config"
-
 // IstioCertsMountPath Istio certs mountpath
 const IstioCertsMountPath = "/etc/istio-certs"
```
davisgomes, line 138: You might be able to delete this since it is used by Prometheus (I would assume exclusively).

pkg/resources/deployments/deployment_test.go
```diff
@@ -5,10 +5,11 @@ package deployments
 
 import (
 	"fmt"
-	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources/nodes"
 	"strings"
 	"testing"
 
+	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources/nodes"
+
```
davisgomes, line 12: Extra space

pkg/constants/constants.go
fredtibbitts: Good catch, removed.

pkg/resources/deployments/deployment_test.go
fredtibbitts: Fixed.

davisgomes (dismissed):

desagar (approved):

davisgomes (approved):

desagar (approved):

PR #210, opened 2022-06-20
WIP: VZ-6150 VMO expects a default storage class even when emptyDir is configured

shashankkumarIITB:
# Description

- Fixes VZ-6150
- VMO expects a default storage class even when emptyDir is configured
- Added logic to ignore PVC creation if defaultStorageClass is not configured or is set to emptyDir

### DIFF
--- a/pkg/vmo/pvc.go
+++ b/pkg/vmo/pvc.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


anders-swanson:
@shashankkumarIITB could you check the unit tests?

pkg/vmo/pvc.go
```diff
@@ -242,5 +242,6 @@ func getDefaultStorageClass(storageClasses []*storagev1.StorageClass) (*storagev
 			return storageClass, nil
 		}
 	}
-	return nil, fmt.Errorf("Failed to find a default storage class")
+	fmt.Println("Failed to find a default storage class. Returning an empty storage class.")
```
anders-swanson, line 245: If you need to emit a log message, please use a logger. We should not be using Println functions

shashankkumarIITB: Done.

anders-swanson (approved):
lgtm

PR #214, opened 2022-06-27
Zabdulre/vz 5822 subtask one

zabdulre:
PR to merge changes made to the go.mod file. Implements requested changes in the previous PR.


stevshap:
anytime go.mod changes, go.sum needs to also

bobdonat:
> anytime go.mod changes, go.sum needs to also

Doesn't look like any packages were added/deleted so I would not expect go.sum to change

stevshap (approved):

PR #218, opened 2022-06-30
VZ 4153: Use ingress class

anders-swanson:


pkg/resources/ingresses/ingress.go
```diff
@@ -16,6 +16,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var defaultIngressClassName = "verrazzano-nginx"
```
fredtibbitts, line 19: Should this be a `const` instead of a `var`?

fredtibbitts (approved):

PR #220, opened 2022-07-14
VZ-6437: Namespace change

theabmitra:
Today velero install happens in a namespace called velero. This could collide with another install of velero if its been installed out of band. 

We will be installing velero in verrazzano-backup namespace.


verrazzano-backup-hook/constants/constants.go
```diff
@@ -9,7 +9,7 @@ const (
 	VerrazzanoSystemNamespace = "verrazzano-system"
 
 	// VerrazzanoNameSpaceName Namespace where Velero components are installed
-	VeleroNameSpace = "velero"
+	VeleroNameSpace = "verrazzano-backup"
```
michael-cico, line 12: should we just use verrazzano-system?

PR #224, opened 2022-07-26
Fix for OS cluster going to yellow state in dev profile when ISM polices are enabled in VZ CR

anajosep:
Fixes VZ-6538 OS cluster goes to yellow state when the hidden indices created by ISM plugin creates with replica=1 as the dev profile has a single node OS cluster. This causes one secondary node to remain assigned causing the cluster to go into yellow state.

### DIFF
--- a/pkg/opensearch/opensearch.go
+++ b/pkg/opensearch/opensearch.go
@@ -21,7 +21,7 @@ type (
 
 const (
 	indexSettings     = `{"index":{"auto_expand_replicas": "0-1"}}`
-	applicationJson   = "application/json"
+	applicationJSON   = "application/json"
 	contentTypeHeader = "Content-Type"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/opensearch/opensearch.go
```diff
@@ -63,6 +63,10 @@ func (o *OSClient) ConfigureISM(vmi *vmcontrollerv1.VerrazzanoMonitoringInstance
 		}
 
 		opensearchEndpoint := resources.GetOpenSearchHTTPEndpoint(vmi)
+		if err := o.updateDefaultIndexSettings(opensearchEndpoint); err != nil {
```
anders-swanson, line 66: Should this be called in the ISM function?
Also, any concern with this being called repeatedly?

anders-swanson (changes requested):
Please fix the linter

anders-swanson (approved):
LGTM

pkg/opensearch/opensearch.go
anajosep: Created a separate Go routine based function. There is no issues if this call is made repeatedly. Only this specific section of the index settings is patched and the rest of the settings are persisted.

PR #227, opened 2022-07-29
VZ-6538 Use index templates to set the 'auto_expand_replicas' to '0-1' for ISM and Jaeger indices in dev profile. 

anajosep:
The earlier fix of using `/_settings` endpoint worked only if the indices were already present. Since it cannot be guaranteed always that the ISM policies are created prior to the VMO initialization, using the index template approach to update the `auto_expand_replicas` of the ISM plugin index `.opendistro_job_scheduler_lock`  to `0-1`

Background of the issue.
-------------------------
In dev profile, with a single node OpenSearch cluster, when an ISM policy is created, the ISM plugin creates an hidden index `.opendistro_job_scheduler_lock` with `replica=1`. This condition cannot be met on a single node cluster and the secondary shard remains un-assigned causing the cluster health to go to yellow (unstable) state. With `auto_expand_replicas` setting, the replica count can be made to 0 for single node clusters and the replica count increments to 1 when additional nodes are added.

### DIFF
--- a/pkg/opensearch/opensearch.go
+++ b/pkg/opensearch/opensearch.go
@@ -125,6 +125,7 @@ func (o *OSClient) SetAutoExpandIndices(vmi *vmcontrollerv1.VerrazzanoMonitoring
 			ch <- fmt.Errorf("expected acknowldegement for index settings update but did not get. Actual response  %v", updatedIndexSettings)
 			return
 		}
+                ch <- nil
 	}()
 
 	return ch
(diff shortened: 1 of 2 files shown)


ayaranga (dismissed):

ayaranga:
As the changes are related to index template, should we be moving this index template creation code to verrazzano OpenSearch component code itself?

ayaranga (approved):

kminder:
This PR needs description content.

anajosep:
> This PR needs description content.

Updated the PR description

kminder (approved):
Approved.

PR #231, opened 2022-08-05
VZ-6699: Have all OpenSearch components use a common node antiaffinity rule

mgianatagh:
Have the affinity rules for es-master, es-data, es-ingest and kibana all have anti affinity rule to spread pods across nodes.

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -317,6 +317,25 @@ func CreateZoneAntiAffinityElement(vmoName string, component string) *corev1.Aff
 	}
 }
 
+// CreateNodeAntiAffinityElement return an Affinity resource for a given VMO instance and component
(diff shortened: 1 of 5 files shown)


stevshap (approved):
LGTM

pkg/resources/helper.go
```diff
@@ -317,6 +317,25 @@ func CreateZoneAntiAffinityElement(vmoName string, component string) *corev1.Aff
 	}
 }
 
+// CreateNodeAntiAffinityElement return an Affinity resource for a given VMO instance and component
```
jmaron99, line 320: this method now exists alongside CreateZoneAntiAffinityElement?  For backwards compatability?

mgianatagh: There is still a case where that method is being used.  I was tempted to remove that use of it too, but was not sure of the implications.  So I left it alone for now.

PR #238, opened 2022-08-16
VZ-6157 - Deploy OpenSearch Dashboard pods one at a time

mgianatagh:
Change install and upgrade of OpenSearch Dashboards to deploy a single pod at a time, to avoid getting errors about indices needing to be migrated.


anders-swanson (approved):
LGTM

PR #244, opened 2022-09-12
VZ 7035, VZ 7034

anders-swanson:
- Don't log err if configmap already exists
- if OS is not Yellow or Green, do not execute the auto expand indices query

### DIFF
--- a/pkg/vmo/configmap.go
+++ b/pkg/vmo/configmap.go
@@ -17,6 +17,7 @@ import (
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources/configmaps"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/vmo/configmap.go
```diff
@@ -42,7 +43,7 @@ func CreateConfigmaps(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMoni
 	// Only create the CM if it doesnt exist. This will allow us to override the provider file e.g. Verrazzano
 	err := createConfigMapIfDoesntExist(controller, vmo, vmo.Spec.Grafana.DashboardsConfigMap, dashboardTemplateMap)
 	if err != nil {
-		controller.log.Errorf("Failed to create dashboard configmap %s: %v", vmo.Spec.Grafana.DashboardsConfigMap, err)
+		controller.log.Debugf("Failed to create dashboard configmap %s: %v", vmo.Spec.Grafana.DashboardsConfigMap, err)
```
bobdonat, line 46: should these errors be buried as debug messages?  doesn't seem like the right thing to do

anders-swanson: I had put them as debug since we were logging the same error multiple times. But I also don't mind to keep them as error. What do you think?

bobdonat: can log.ErrorfThrottledNewErr be used

anders-swanson: cleaned up errors

PR #249, opened 2022-09-14
VZ-7002: Prevent logging of error messages due to Opensearch not being ready

lake-of-dreams:
Check for Opensearch pods to be ready before accessing in `ConfigureISM` , `MigrateOldIndices` and `SetAutoExpandIndices` to prevent logging error message due to pods not being up yet

### DIFF
--- a/pkg/opensearch/opensearch.go
+++ b/pkg/opensearch/opensearch.go
@@ -7,16 +7,22 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/http"
+
 	vmcontrollerv1 "github.com/verrazzano/verrazzano-monitoring-operator/pkg/apis/vmcontroller/v1"
+	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/constants"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources"
 	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/resources/nodes"
-	"net/http"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/labels"
+	appslistersv1 "k8s.io/client-go/listers/apps/v1"
 )
 
 type (
 	OSClient struct {
-		httpClient *http.Client
-		DoHTTP     func(request *http.Request) (*http.Response, error)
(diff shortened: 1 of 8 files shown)


pkg/opensearch/opensearch.go
```diff
@@ -130,3 +149,25 @@ func (o *OSClient) SetAutoExpandIndices(vmi *vmcontrollerv1.VerrazzanoMonitoring
 
 	return ch
 }
+
+//IsOpenSearchReady returns true when all OpenSearch pods are ready, false otherwise
+func (o *OSClient) IsOpenSearchReady(vmi *vmcontrollerv1.VerrazzanoMonitoringInstance) bool {
```
anders-swanson, line 154: we have IsGreen method, would that work instead of creating a new method?

lake-of-dreams: The jira success criteria/solution is 
```The OpenSearch deployment should be checked for readiness before these queries occur so that no confusing error output is produced.```

When I look at [getOpenSearchClusterHealth](https://github.com/verrazzano/verrazzano-monitoring-operator/blob/master/pkg/opensearch/health.go#L124) which is being called from `IsGreen` , it is actually invoking [/_cluster/health](https://opensearch.org/docs/latest/opensearch/rest-api/cluster-health/) which seems to be more than just checking that Deployment .

Now since the functions in question (`ConfigureISM` , `MigrateOldIndices` and `SetAutoExpandIndices`), did not wait for the `IsGreen` status before, my question is 

1. Is it ok to processing in these functions to continue when the cluster status is not `Green` but pods are ready? 
2. Does it have to be `Green` or `Yellow` will also do?
3. Should the `IsOpenSearchReady` be kept as is and in `getOpenSearchClusterHealth` call this before making the api call?

anders-swanson: We had initially put no 'healthy' check in these functions, because we thought: "eventually query will succeed once OS comes online, so there is no point in putting another check in front of it". 
And yes, IsGreen may be too restrictive of a check, your check is probably more in line with what we want. We may also consider squashing these error messages unless the error is actually from OpenSearch server. These are kind of background routines anyway.

anders-swanson (approved):

PR #253, opened 2022-09-20
VZ-7099: Fix bug that causes deployments to not be deleted

fredtibbitts:
This PR fixes a bug that occurs in the following scenario:
1. User upgrades Verrazzano installation to 1.4
2. After the VMO is upgraded, attempting to create/update deployments fails because the OpenSearch status is yellow
3. Because create/update deployments fails, it never gets to the point of processing deployment deletions

As a result, the old VMI system prometheus deployment does not always get deleted when it should. To fix the problem, I separated out the delete portion of the CreateDeployments function into a new function. The controller calls this new function whether CreateDeployments fails or not.

I have no idea why the OpenSearch health does not report green in some cases, but this fixes the problem of deployments not being deleted.

### DIFF
--- a/pkg/vmo/controller.go
+++ b/pkg/vmo/controller.go
@@ -536,22 +536,23 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
 		c.log.Errorf("Failed to create Deployment specs for VMI %s: %v", vmo.Name, err)
 		functionMetric.IncError()
 		errorObserved = true
-	}
-	deploymentsDirty, err = CreateDeployments(c, vmo, expectedDeployments, existingCluster)
-	if err != nil {
-		c.log.Errorf("Failed to create/update deployments for VMI %s: %v", vmo.Name, err)
-		functionMetric.IncError()
-		errorObserved = true
-	}
-	// OSD is a special case, so add it to the expected deployments (if it should exist)
-	osd := deployments.NewOpenSearchDashboardsDeployment(vmo)
-	if osd != nil {
-		expectedDeployments.Deployments = append(expectedDeployments.Deployments, osd)
-	}
-	if err = DeleteDeployments(c, vmo, expectedDeployments); err != nil {
-		c.log.Errorf("Failed to delete deployments for VMI %s: %v", vmo.Name, err)
-		functionMetric.IncError()
-		errorObserved = true
+	} else {
+		deploymentsDirty, err = CreateDeployments(c, vmo, expectedDeployments, existingCluster)
+		if err != nil {
+			c.log.Errorf("Failed to create/update deployments for VMI %s: %v", vmo.Name, err)
+			functionMetric.IncError()
+			errorObserved = true
+		}
+		// OSD is a special case, so add it to the expected deployments (if it should exist)
+		osd := deployments.NewOpenSearchDashboardsDeployment(vmo)
+		if osd != nil {
+			expectedDeployments.Deployments = append(expectedDeployments.Deployments, osd)
+		}
+		if err = DeleteDeployments(c, vmo, expectedDeployments); err != nil {
+			c.log.Errorf("Failed to delete deployments for VMI %s: %v", vmo.Name, err)
+			functionMetric.IncError()
+			errorObserved = true
+		}
 	}
 
 	/*********************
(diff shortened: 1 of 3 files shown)


pkg/vmo/controller.go
```diff
@@ -527,17 +528,24 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
...
 	if !errorObserved {
-		deploymentsDirty, err = CreateDeployments(c, vmo, pvcToAdMap, existingCluster)
+		deploymentsDirty, expectedDeployments, err = CreateDeployments(c, vmo, pvcToAdMap, existingCluster)
 		if err != nil {
 			c.log.Errorf("Failed to create/update deployments for VMI %s: %v", vmo.Name, err)
 			functionMetric.IncError()
 			errorObserved = true
 		}
 	}
+	if err = DeleteDeployments(c, vmo, expectedDeployments); err != nil {
```
pfmackin, line 543: if errorObserved == nil then expectedDeployments will be nil also.  Is that what you want?

fredtibbitts: errorObserved is a bool. expectedDeployments may be non-nil even if errorObserved is true.

fredtibbitts: I made more changes to hopefully improve this. I moved the creation of the expected deployments out of the CreateDeployments function and I pass it into both CreateDeployments and DeleteDeployments. This code is a bit of mess, and pulling it apart has been challenging.

pkg/vmo/controller.go
```diff
@@ -527,17 +528,33 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
 	}
 
 	/*********************
-	 * Create Deployments
+	 * Create, update, and delete Deployments
 	 **********************/
 	var deploymentsDirty bool
-	if !errorObserved {
-		deploymentsDirty, err = CreateDeployments(c, vmo, pvcToAdMap, existingCluster)
+	expectedDeployments, err := deployments.New(vmo, c.kubeclientset, c.operatorConfig, pvcToAdMap)
```
pfmackin, line 534: Are these deployments from the old install?

pkg/vmo/controller.go
```diff
@@ -527,17 +528,33 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
...
+	if err != nil {
+		c.log.Errorf("Failed to create Deployment specs for VMI %s: %v", vmo.Name, err)
+		functionMetric.IncError()
+		errorObserved = true
+	} else {
+		deploymentsDirty, err = CreateDeployments(c, vmo, expectedDeployments, existingCluster)
```
pfmackin, line 540: why the name deploymentsDirty?  Are these deployments for the new install?

pkg/vmo/controller.go
```diff
@@ -527,17 +528,33 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
...
 			c.log.Errorf("Failed to create/update deployments for VMI %s: %v", vmo.Name, err)
 			functionMetric.IncError()
 			errorObserved = true
 		}
+		// OSD is a special case, so add it to the expected deployments (if it should exist)
+		osd := deployments.NewOpenSearchDashboardsDeployment(vmo)
+		if osd != nil {
+			expectedDeployments.Deployments = append(expectedDeployments.Deployments, osd)
+		}
+		if err = DeleteDeployments(c, vmo, expectedDeployments); err != nil {
```
pfmackin, line 551: Not clear why expectedDeployments are being deleted.  Are these the deployments from the old install?

pkg/vmo/controller.go
fredtibbitts: No, these are the deployments that are expected to exist now. The `CreateOrUpdateDeployments` and `DeleteDeployments` functions reconcile to make the actual deployments match the expected deployments. I added a comment.

pkg/vmo/controller.go
fredtibbitts: Renamed the function to `CreateOrUpdateDeployments`.

pkg/vmo/controller.go
fredtibbitts: The expected deployments are passed in so they can be compared with the actual deployments, and any deployments that are not expected are deleted. I added a comment to try to clarify.

pfmackin (approved):

PR #254, opened 2022-09-23
VZ-5615 Update the Verrazzano user to Admin in Grafana

davisgomes:
This PR enables the Verrazzano user to become a Grafana admin if requested. To request the verrazzano admin user, the controller looks for this annotation on the VMO:
`grafana-admin-update: true`

This process acts as a state machine that is determined based on the status of the Grafana deployment. The state should be directed (but not acyclic) and can be entered at any stage (this is required if the Grafana deployment or the VMO were to crash in the middle of this process):

- **Setup**
  - If basic auth is disabled and the authproxy is enabled _without_ the `grafana-admin-update: completed` annotation on the Grafana deployment, we enter setup mode
  - This mode will enable basic authentication to enable us to use the Grafana Admin API which we will use in the next step
  - This requires a pod restart for Grafana
  - if successful, transition to the **Request** state
- **Request**
  - the request state does either both or the second of these two items
  - if the verrazzano user does not exist (the console was not accessed before upgrade and the user was never created)
    - create the user and maintain the **Request** stage
  - if the verrazzano user exists, we request that user have Grafana admin permissions
  - if this is successful, we transition to the **Complete** stage by labeling the Grafana deployment with `grafana-admin-update: completed`
- **Complete**
  - if the `grafana-admin-update: completed` is set on the Grafana deployment
    - we return to the original Grafan configuration (authproxy enabled, basic auth disabled)
  - we maintain this state by keeping the annotation `grafana-admin-update: completed`

This is all done in one process (it's pretty quick, less than a minute). I fear that continually reconciling (which would be ideal) would cause cascading changes in the VMO and make this change much larger than it needs to be. I think that because this is a one-time process (it should only run once in the lifecycle of Verrazzano) that we can get away with running this continuously.

### DIFF
--- a/pkg/vmo/deployment.go
+++ b/pkg/vmo/deployment.go
@@ -230,31 +230,6 @@ func determineGrafanaState(controller *Controller, deployment *appsv1.Deployment
 			return Complete, nil
 		}
 	}
-	// if the Verrazzano user is not found, we need to create one
-	// this occurs if the console is not accessed before this process is run
-	if grafanaResponse.StatusCode == 404 {
-		vzUserReg := grafanaUserRegistration{
-			Name:  "",
-			Email: "verrazzano",
-			Login: "verrazzano",
-			// because we use an auth proxy to authenticate, we do not need to supply a valid password here
-			// This field is required, so it must be populated
-			Password: "verrazzano",
-		}
-		requestData, err := json.Marshal(&vzUserReg)
-		if err != nil {
-			controller.log.Errorf("Failed to encode the request to create the Verrazzano user: %v", err)
-			return 0, err
-		}
-		grafanaURL.Path = "api/admin/users"
-		grafanaURL.RawQuery = ""
-		grafanaURL.User = url.UserPassword("admin", "admin")
-		registerResponse, err := http.Post(grafanaURL.String(), "application/json", bytes.NewBuffer(requestData))
-		if err != nil || registerResponse.StatusCode != 200 {
-			controller.log.Errorf("Failed to request the Verrazzano user creation, status %d: %v", registerResponse.StatusCode, err)
-			return 0, err
-		}
-	}
 	if grafanaResponse.StatusCode == 503 {
 		controller.log.Progressf("Waiting for Grafana pod to be ready before request, status: %d", grafanaResponse.StatusCode)
 		return Setup, nil
@@ -277,10 +252,37 @@ func requestGrafanaAdmin(controller *Controller, grafanaURL url.URL) error {
 	grafanaURL.Path = "api/users/lookup"
 	grafanaURL.RawQuery = "loginOrEmail=verrazzano"
 	grafanaResponse, err := http.Get(grafanaURL.String())
-	if err != nil || grafanaResponse.StatusCode != 200 {
+	if err != nil && grafanaResponse.StatusCode != 404 {
 		controller.log.Errorf("Failed to get Verrazzano user information from Grafana with request %s, status %d: %v", grafanaURL.String(), grafanaResponse.StatusCode, err)
 		return err
 	}
+	// if the Verrazzano user is not found, we need to create one
+	// this occurs if the console is not accessed before this process is run
+	if grafanaResponse.StatusCode == 404 {
+		controller.log.Once("Failed to find Verrazzano user in Grafana, creating now")
+		vzUserReg := grafanaUserRegistration{
+			Name:  "",
+			Email: "verrazzano",
+			Login: "verrazzano",
+			// because we use an auth proxy to authenticate, we do not need to supply a valid password here
+			// This field is required, so it must be populated
+			Password: "verrazzano",
+		}
+		requestData, errReg := json.Marshal(&vzUserReg)
+		if errReg != nil {
+			controller.log.Errorf("Failed to encode the request to create the Verrazzano user: %v", errReg)
+			return errReg
+		}
+		grafanaURL.Path = "api/admin/users"
+		grafanaURL.RawQuery = ""
+		registerResponse, errReg := http.Post(grafanaURL.String(), "application/json", bytes.NewBuffer(requestData))
+		if errReg != nil || registerResponse.StatusCode != 200 {
+			controller.log.Errorf("Failed to request the Verrazzano user creation, status %d: %v", registerResponse.StatusCode, errReg)
+			return err
+		}
+		return err
+	}
+
 	var vzUserInfo grafanaUserInfo
 	err = json.NewDecoder(grafanaResponse.Body).Decode(&vzUserInfo)
 	if err != nil {
--- a/pkg/vmo/deployment.go
+++ b/pkg/vmo/deployment.go
@@ -69,7 +69,7 @@ type grafanaAdminRequest struct {
 }
 
 // grafanaAdminResponse is the expected response body from a Grafana admin permissions request
-type grafanaAdminResponse struct {
+type grafanaMessageResponse struct {
 	Message string `json:"message"`
 }
 
@@ -211,8 +211,17 @@ func determineGrafanaState(controller *Controller, deployment *appsv1.Deployment
 		controller.log.Errorf("Failed to get Verrazzano user information from Grafana with request %s: %v", grafanaURL.String(), err)
 		return 0, err
 	}
-	if grafanaResponse.Status == "200" {
+	if grafanaResponse.StatusCode == 200 {
 		return Complete, nil
+	} else {
+		// We expect a message response for a
+		var messageResponse grafanaMessageResponse
+		err = json.NewDecoder(grafanaResponse.Body).Decode(&messageResponse)
+		if err != nil {
+			controller.log.Errorf("Failed to decode the response body of the incomplete request with status %d: %v", grafanaResponse.StatusCode, err)
+			return 0, err
+		}
+		controller.log.Progressf("Request to get Verrazzano user info from from the Grafana pod was unsuccessful status: %d message: %s", grafanaResponse.StatusCode, messageResponse.Message)
 	}
 
 	// Check the deployment pod env vars to determine if basic auth is enabled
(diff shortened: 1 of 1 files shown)


pkg/vmo/deployment.go
```diff
@@ -16,12 +22,64 @@ import (
...
 
+const (
+	grafanaAdminAnnotation = "grafana-admin-update"
+	completeVal            = "complete"
+	authProxyEnvName       = "GF_AUTH_PROXY_ENABLED"
+	basicAuthEnvName       = "GF_AUTH_BASIC_ENABLED"
+)
+
+// GrafanaAdminState Grafana Admin Update state enum
+type grafanaAdminState int
```
mbenmclean, line 40: Is there a reason why this is an int instead of a string?

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+	// this occurs if the console is not accessed before this process is run
+	if grafanaResponse.StatusCode == 404 {
+		controller.log.Once("Failed to find Verrazzano user in Grafana, creating now")
+		vzUserReg := grafanaUserRegistration{
+			Name:  "",
+			Email: "verrazzano",
+			Login: "verrazzano",
+			// because we use an auth proxy to authenticate, we do not need to supply a valid password here
+			// This field is required, so it must be populated
+			Password: "verrazzano",
```
mbenmclean, line 251: We have generated a Grafana-admin secret for this value. We should probably use that. This password is probably a security problem.

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+		controller.log.Errorf("Failed to get Verrazzano user information from Grafana with request %s, status %d: %v", grafanaURL.String(), grafanaResponse.StatusCode, err)
+		return true, err
+	}
+	// if the Verrazzano user is not found, we need to create one
+	// this occurs if the console is not accessed before this process is run
+	if grafanaResponse.StatusCode == 404 {
+		controller.log.Once("Failed to find Verrazzano user in Grafana, creating now")
+		vzUserReg := grafanaUserRegistration{
+			Name:  "",
+			Email: "verrazzano",
```
mbenmclean, line 247: Does the email need to be not empty?

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+			Login: "verrazzano",
```
mbenmclean, line 248: This is set up in the Grafana-admin secret to be the same value as the vmi-secret. Should probably make a constant for that and set both values to it so they don't diverge

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+			// because we use an auth proxy to authenticate, we do not need to supply a valid password here
+			// This field is required, so it must be populated
+			Password: "verrazzano",
+		}
+		requestData, errReg := json.Marshal(&vzUserReg)
+		if errReg != nil {
+			controller.log.Errorf("Failed to encode the request to create the Verrazzano user: %v", errReg)
```
mbenmclean, line 255: "Verrazzano Grafana user"
So they know the context of the log. Make sure all these logs reference that you are updating the Grafana user so if something goes wrong the user knows that.

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+			return true, errReg
+		}
+		grafanaURL.Path = "api/admin/users"
+		grafanaURL.RawQuery = ""
+		registerResponse, errReg := http.Post(grafanaURL.String(), "application/json", bytes.NewBuffer(requestData))
+		if errReg != nil || registerResponse.StatusCode != 200 {
+			controller.log.Errorf("Failed to request the Verrazzano user creation, status %d: %v", registerResponse.StatusCode, errReg)
```
mbenmclean, line 262: reference grafana

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+	grafanaRequest, err := http.NewRequest(http.MethodPut, grafanaURL.String(), bytes.NewBuffer(requestData))
+	if err != nil {
+		controller.log.Errorf("Failed to create request for admin permissions for the Verrazzano user: %v", err)
+		return true, err
+	}
+	grafanaRequest.Header.Set("Content-type", "application/json")
+	client := http.Client{}
+	grafanaResponse, err = client.Do(grafanaRequest)
+	if err != nil || grafanaResponse.StatusCode != 200 {
+		controller.log.Errorf("Failed to request admin permissions for the Verrazzano user, status %d: %v", grafanaResponse.StatusCode, err)
```
mbenmclean, line 293: reference grafana

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,180 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+		return true, err
+	}
+
+	// Verify the response gives a valid response message
+	var adminUserResponse grafanaMessageResponse
+	err = json.NewDecoder(grafanaResponse.Body).Decode(&adminUserResponse)
+	if err != nil {
+		controller.log.Errorf("Failed to decode the response body of the Verrazzano admin request: %v", err)
```
mbenmclean, line 301: reference grafana

pkg/vmo/deployment.go
```diff
@@ -212,7 +460,9 @@ func updateDeployment(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMoni
 	curDeployment.Spec.Selector = existingDeployment.Spec.Selector
 	specDiffs := diff.Diff(existingDeployment, curDeployment)
 	if specDiffs != "" {
-		controller.log.Oncef("Deployment %s/%s has spec differences %s", curDeployment.Namespace, curDeployment.Name, specDiffs)
+		if logDiff {
```
mbenmclean, line 463: when is this true and when is this false? Is there a reason why you don't want to log diffs in some cases now. Previously we were logging all diffs.

pkg/vmo/deployment.go
davisgomes: Yea you're right. I was trying to avoid another resource grab, but this is probably a security problem at some level even if basic auth isn't enabled. I'll update it.

pkg/vmo/deployment.go
davisgomes: This was how the user was generated from the authproxy, so I did not want to diverge from that user info if I did not have to.

pkg/vmo/deployment.go
davisgomes: This caused unnecessary diff dumps that are not important for the user or related to user changes. Because we are controlling the diff changes, I decided not to log it. I log it for all existing deployment updates.

pkg/vmo/deployment.go
davisgomes: This acts as an enum, so I'm using an empty integer value to store these values.  Here's where I referenced this idea from: https://www.sohamkamani.com/golang/enums/

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,192 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
 	return nil
 }
 
+// updateGrafanaAdminUser updates the Grafana deployment to make the Verrazzano user a Grafana Admin
+func updateGrafanaAdminUser(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, grafanaDeployment *appsv1.Deployment, curDeployment *appsv1.Deployment) (bool, error) {
+	grafanaURL := url.URL{
+		Scheme: "HTTP",
+		Host:   fmt.Sprintf("%s.%s.svc.cluster.local:3000", resources.GetMetaName(vmo.Name, config.Grafana.Name), vmo.Namespace),
```
pfmackin, line 142: Is this port always 3000?  Do we hard code this in other places?

pkg/vmo/deployment.go
```diff
@@ -77,6 +135,192 @@ func updateOpenSearchDashboardsDeployment(osd *appsv1.Deployment, controller *Co
...
+	}
+	controller.log.Progress("Verrazzano user has not yet been upgraded to admin in Grafana")
+
+	// Check the deployment pod env vars to determine if basic auth is enabled
+	// If so, we should enter the request state
+	// If not, we should enter the setup state
+	// Collect necessary information from the Grafana pod env vars
+	basicAuthEnabled := false
+	authProxyEnabled := false
+	for _, envVar := range deployment.Spec.Template.Spec.Containers[0].Env {
```
pfmackin, line 217: Do you expect to have different results for different pods?

pkg/vmo/deployment.go
```diff
@@ -212,7 +472,9 @@ func updateDeployment(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMoni
 	curDeployment.Spec.Selector = existingDeployment.Spec.Selector
 	specDiffs := diff.Diff(existingDeployment, curDeployment)
 	if specDiffs != "" {
-		controller.log.Oncef("Deployment %s/%s has spec differences %s", curDeployment.Namespace, curDeployment.Name, specDiffs)
+		if logDiff {
```
pfmackin, line 475: We don't normally log diffs, how much output will this generate and why are we logging it?

davisgomes: I disabled the diff logging for this change and only maintained the logging that was existing. It was causing large and unhelpful log messages so I filtered it out for all the changes I made. The only diffs that are logged are for existing deployment updates.

pkg/vmo/deployment.go
davisgomes: Yes, there are a few possible deployment generations for Grafana. This takes care of all possible env var arrangements for Grafana and only returns true if the right configuration exists.

PR #255, opened 2022-09-28
Enhance logging.

IshaGirdhar:
Introduce and use VZ logger ErrorfThrottled method to reduce the error volume for errors.

### DIFF
--- a/pkg/util/logs/vzlog/vzlog.go
+++ b/pkg/util/logs/vzlog/vzlog.go
@@ -79,6 +79,12 @@ type ProgressLogger interface {
 	// ErrorfNewErr formats an error, logs it, then returns the formatted error
 	ErrorfNewErr(template string, args ...interface{}) error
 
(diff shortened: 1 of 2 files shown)


pkg/util/logs/vzlog/vzlog.go
```diff
@@ -225,9 +231,10 @@ func (c *LogContext) EnsureLogger(key string, sLogger SugaredLogger, zap *zap.Su
 	log, ok := c.loggerMap[key]
 	if !ok {
 		log = &verrazzanoLogger{
-			context:       c,
-			frequencySecs: 60,
-			trashMessages: make(map[string]bool),
+			context:         c,
+			frequencySecs:   60,
+			trashMessages:   make(map[string]bool),
```
anders-swanson, line 236: any concurrency concerns with these maps? afaik logger may be used by multiple goroutines, but im not sure.

IshaGirdhar: yeah took care of it by adding mutex.

anders-swanson (approved):

PR #257, opened 2022-09-30
Sukruthbhat/vz 7157

SukruthBhat:
Updated development.md by comparing it with the updated Makefile.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -81,10 +81,21 @@ kubectl apply -f k8s/manifests/verrazzano-monitoring-operator.yaml
 
 ## Running unit and integration tests
 
+To perform static analysis:
+
+```
+make golangci-lint
+```
 To run unit tests:
 
 ```
 make unit-test
+```
+
+To perform both static analysis and run unit test in one go:
+
+```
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -81,12 +81,23 @@ kubectl apply -f k8s/manifests/verrazzano-monitoring-operator.yaml
...
 To run unit tests:
 
 ```
 make unit-test
 ```
 
+To perform both static analysis and run unit test in one go:
+
+```
+make check
```
gilbode, line 98: Last I checked this only runs the linters though the behavior described here sounds better

SukruthBhat: Should I remove it then??

SukruthBhat: .PHONY: check
check: golangci-lint unit-test
Target command check was calling both the linter target command and unit test command..that's why I used that description.

gilbode: My mistake - thought this was verrazzano repo not verrazzano-monitoring-operator.

gilbode (approved):

PR #270, opened 2022-10-13
VZ-7330:Successful wrongly spelt

SukruthBhat:
When misspell linter was added to Verrazzano, Jenkins build was failing in the System component metrics stage. After probing I figured out that the failure was because vmoTimestampMetric, which is one of the constants for sample metrics of system components had it's value changed from "vmo_configmap_last_succesful_timestamp" to "vmo_configmap_last_successful_timestamp" as successful was reported as a spelling error by misspell linter. These errors traced back to metrics exporter.go file of Verrazzano Monitoring Operator repository.

### DIFF
--- a/pkg/metricsexporter/metricsexporter.go
+++ b/pkg/metricsexporter/metricsexporter.go
@@ -356,7 +356,7 @@ func (md *metricsDelegate) initializeFailedMetricsArray() {
 	}
 }
 
-// registerMetricsHandlersHelper loops through the failed metrics map and deletes metrics which have been registered succesSfully
+// registerMetricsHandlersHelper loops through the failed metrics map and deletes metrics which have been registered successfully
 func (md *metricsDelegate) registerMetricsHandlersHelper() error {
 	var errorObserved error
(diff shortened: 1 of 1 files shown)


pkg/metricsexporter/metricsexporter.go
```diff
@@ -356,7 +356,7 @@ func (md *metricsDelegate) initializeFailedMetricsArray() {
 	}
 }
 
-// registerMetricsHandlersHelper loops through the failed metrics map and deletes metrics which have been registered succesfully
+// registerMetricsHandlersHelper loops through the failed metrics map and deletes metrics which have been registered succesSfully
```
adiforluls, line 359: Nit: successfully and not succesSfully

adiforluls (approved):

pkg/metricsexporter/metricsexporter.go
SukruthBhat: oh yeah my bad.

adiforluls (approved):

PR #278, opened 2022-11-02
VZ-7396: Remove ES access URL on a new install and preserve it for upgrade. 

Abhijithang:
PR does the following:
- New install - Removes Elasticsearch/Kibana references from ingresses.
- Upgrade - preserves the deprecated Elasticsearch/Kibana ingress URL by updating the new ingresses with an additional Rule and TLS host.

### DIFF
--- a/pkg/resources/ingresses/ingress.go
+++ b/pkg/resources/ingresses/ingress.go
@@ -250,7 +250,7 @@ func getIngressRuleForESHost(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, c
 	}
 	pathType := netv1.PathTypeImplementationSpecific
 	serviceName := resources.AuthProxyMetaName()
-	ingressHostES := resources.OidcProxyIngressHostES(vmo, component)
+	ingressHostES := resources.OidcProxyIngressHost(vmo, component)
 	ingressRule := netv1.IngressRule{
 		Host: ingressHostES,
 		IngressRuleValue: netv1.IngressRuleValue{
--- a/pkg/resources/ingresses/ingress.go
+++ b/pkg/resources/ingresses/ingress.go
@@ -167,7 +167,7 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) ([]*netv1.Ingress, er
 		if config.ElasticsearchIngest.OidcProxy != nil {
 			ingress := newOidcProxyIngress(vmo, &config.ElasticsearchIngest)
 			ingress.Annotations["nginx.ingress.kubernetes.io/proxy-body-size"] = "65M"
-			ingressES := newOidcProxyIngressES(vmo, &config.ElasticsearchIngest)
+			ingressES := newOidcProxyIngressOS(vmo, &config.ElasticsearchIngest)
 			ingresses = append(ingresses, ingress)
 			ingressES.Annotations["nginx.ingress.kubernetes.io/proxy-body-size"] = "65M"
 			ingresses = append(ingresses, ingressES)
@@ -296,14 +296,13 @@ func newOidcProxyIngress(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, compo
 }
 
 // newOidcProxyIngress creates the Ingress of the OidcProxy
-func newOidcProxyIngressES(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, component *config.ComponentDetails) *netv1.Ingress {
+func newOidcProxyIngressOS(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, component *config.ComponentDetails) *netv1.Ingress {
 	port, err := strconv.ParseInt(resources.AuthProxyPort(), 10, 32)
 	if err != nil {
 		port = 8775
 	}
 	serviceName := resources.AuthProxyMetaName()
-	//ingressHost := resources.OidcProxyIngressHost(vmo, component)
-	ingressHost := resources.OidcProxyIngressHostES(vmo, component)
+	ingressHost := resources.OidcProxyIngressHostOS(vmo, component)
 	pathType := netv1.PathTypeImplementationSpecific
 	ingressClassName := getIngressClassName(vmo)
 	ingressRule := netv1.IngressRule{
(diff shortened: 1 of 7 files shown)


pkg/resources/ingresses/ingress.go
```diff
@@ -204,6 +203,45 @@ func noAuthOnHealthCheckSnippet(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance
 `
 }
 
+func setIngressTLSHostES(ingressHost string, ingressTLS []netv1.IngressTLS) []netv1.IngressTLS {
+	ingressTLS[0].Hosts = append(ingressTLS[0].Hosts, ingressHost)
+	return ingressTLS
+}
+
+func getIngressRuleForESHost(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, component *config.ComponentDetails) netv1.IngressRule {
+
```
anders-swanson, line 212: extra newline

pkg/vmo/ingress.go
```diff
@@ -51,10 +60,21 @@ func CreateIngresses(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonit
...
-
 		controller.log.Debugf("Applying Ingress '%s' in namespace '%s' for VMI '%s'\n", ingName, vmo.Namespace, vmo.Name)
 		existingIngress, err := controller.ingressLister.Ingresses(vmo.Namespace).Get(ingName)
 		if existingIngress != nil {
+			if existingIngress.Name == "vmi-system-os-ingest" {
+				if DoesIngressContainDeprecatedESHost(existingIngress, vmo, &config.ElasticsearchIngest) {
+					curIngress = ingresses.AddNewRuleAndHostTLSForIngress(vmo, curIngress, &config.ElasticsearchIngest)
+				}
+				OSIngest = curIngress
+			} else if existingIngress.Name == "vmi-system-opensearchdashboards" {
```
anders-swanson, line 71: On a post-upgrade build, wont the existing ingress be named `vmi-<VMI Name>-kibana` or similar for elasticsearch?

Abhijithang: For new installation -creating a new ingress with vmi-system-opensearchdashboard and vmi-system-opensearch names. I am not aware of the requirement for removal of "system" from ingress, if there is any such requirement please let me know, I will make the changes accordingly. So here the goal is to eliminate elasticsearch and kibana references.

Abhijithang: For upgrade - Create a new ingress but update the hosts with deprecated ES/Kibana urls so that users who bookmarked it could still access it.

pkg/vmo/ingress.go
```diff
@@ -85,6 +105,27 @@ func CreateIngresses(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonit
 	}
 	for _, ingress := range existingIngressList {
 		if !contains(ingressNames, ingress.Name) {
+			// For upgrade check, if the user has deprecated Elasticsearch/Kibana ingress
+			// Then update the new opensearch/opensearchdashboards ingress with additional Elasticsearch/Kibana rule and hosts
+			// To support access to the deprecated Elasticsearch/Kibana URL. This case is only for upgrade.
+			if ingress.Name == "vmi-system-es-ingest" && OSIngest != nil {
```
anders-swanson, line 111: It would be good to combine this with the previous block. The OS Ingest and OSD Ingresses need the deprecated ingress rules added IF:
1. The existing OS/OSD ingresses have the hostname
2. es-ingest or kibana ingresses exist (they should be deleted, afterwards, right?)

pkg/vmo/ingress.go
Abhijithang: Sorry, I think I got what you were trying to say. I will update the code with VMI name. My bad.

pkg/vmo/ingress.go
anders-swanson: Could we coalesce the OS/OSD ingress mutate to a common area above these loops? I.e., create a function like `addDeprecatedHostsIfNecessary`. This is mostly a style thing, so do not take it as necessary.

pkg/vmo/ingress.go
```diff
@@ -143,12 +105,33 @@ func CreateIngresses(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonit
...
 }
 
-// DoesIngressContainDeprecatedESHost Checks if the ingress contain Deprecated ES host
-func DoesIngressContainDeprecatedESHost(ingress *netv1.Ingress, vmo *vmcontrollerv1.VerrazzanoMonitoringInstance, componentDetails *config.ComponentDetails) bool {
-	for _, rule := range ingress.Spec.Rules {
-		if rule.Host == resources.OidcProxyIngressHost(vmo, componentDetails) {
-			return true
-		}
+// getRequiredExistingIngresses retrieves the required ingress objects
+func getRequiredExistingIngresses(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) map[string]*netv1.Ingress {
```
anders-swanson, line 109: Can we use this codes (already exists to get existing ingresses as a list):
```
	selector := labels.SelectorFromSet(map[string]string{constants.VMOLabel: vmo.Name})
	existingIngressList, err := controller.ingressLister.Ingresses(vmo.Namespace).List(selector)
```

And then repurpose this function to turn the list into a map? That would optimize the number of calls to the API server.

pkg/resources/ingresses/ingress.go
Abhijithang: Done

pkg/vmo/ingress.go
```diff
@@ -106,32 +106,30 @@ func CreateIngresses(controller *Controller, vmo *vmcontrollerv1.VerrazzanoMonit
...
-	if ingressObject != nil {
-		existingIngressMap[ingressName] = ingressObject
-	}
-	// Get Opensearchdashboards ingress object
-	ingressName = resources.GetMetaName(vmo.Name, config.OpenSearchDashboards.Name)
-	ingressObject, _ = controller.ingressLister.Ingresses(vmo.Namespace).Get(ingressName)
-	if ingressObject != nil {
-		existingIngressMap[ingressName] = ingressObject
+	for _, existingIngress := range existingIngressList {
+		// Get Elasticsearch ingress object
```
anders-swanson, line 113: can simplify to just `existingIngressMap[existingIngress.Name] = existingIngress`? Any reason to filter for only these ingresses?

Abhijithang: No compelling reason. Just added the required information to the maps. Either way it should be good. I can make the above change.

Abhijithang: Done

anders-swanson (dismissed):

anders-swanson (approved):

PR #282, opened 2022-11-09
VZ-7585:  verrazzano-monitoring-operator make manifests/make generate does not run in pipeline 

pshirodkar-orcl:
Hardcoding `sigs.k8s.io/controller-tools` to `v0.8.0` since it does not exist as a module dependency in `go.mod`.
Make sure to correctly populate `$GOBIN` environment variable.
Switching from `command` bash shell builtin to `which` since `command` is hashed command. If someone removes `controller-gen` binary from their machine, then `command -v controller-gen` can still falsely indicate the `controller-gen` binary exists.
Added make target for repo clean for manifests verification.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,6 +64,21 @@ pipeline {
             }
         }
 
+        stage('Check Repo Clean') {
+            steps {
+                checkRepoClean()
+            }
+            post {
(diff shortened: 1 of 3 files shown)


anders-swanson (approved):
LGTM

PR #285, opened 2022-11-17
VZ-7396: Remove ES access URL on a new install and preserve it for upgrade.

Abhijithang:
This PR does the following:
- New install - Removes Elasticsearch/Kibana references from ingresses.
- Upgrade - preserves the deprecated Elasticsearch/Kibana ingress URL access by creating a redirect ingress for permanent redirection.

### DIFF
--- a/pkg/config/components.go
+++ b/pkg/config/components.go
@@ -133,9 +133,8 @@ var OidcProxy = ComponentDetails{
 
 // ElasticsearchIngest is the default Elasticsearch IngestNodes configuration
 var ElasticsearchIngest = ComponentDetails{
-	Name:         "es-ingest",
-	EndpointName: "elasticsearch",
-	//NOTE: update ELASTICSEARCH_WAIT_TARGET_VERSION env (constants.ESWaitTargetVersionEnv) value to match the version reported by this image
+	Name:              "es-ingest",
+	EndpointName:      "elasticsearch",
 	EnvName:           "ELASTICSEARCH_IMAGE",
 	ImagePullPolicy:   constants.DefaultImagePullPolicy,
 	Port:              constants.OSHTTPPort,
@@ -147,9 +146,8 @@ var ElasticsearchIngest = ComponentDetails{
 
 // OpensearchIngest is the default Opensearch IngestNodes configuration
 var OpensearchIngest = ComponentDetails{
-	Name:         "os-ingest",
-	EndpointName: "opensearch",
-	//NOTE: update ELASTICSEARCH_WAIT_TARGET_VERSION env (constants.ESWaitTargetVersionEnv) value to match the version reported by this image
+	Name:              "os-ingest",
+	EndpointName:      "opensearch",
 	EnvName:           "ELASTICSEARCH_IMAGE",
 	ImagePullPolicy:   constants.DefaultImagePullPolicy,
 	Port:              constants.OSHTTPPort,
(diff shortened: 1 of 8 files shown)


pkg/config/components.go
```diff
@@ -120,6 +145,34 @@ var ElasticsearchIngest = ComponentDetails{
 	OidcProxy:         &OidcProxy,
 }
 
+// OpensearchIngest is the default Opensearch IngestNodes configuration
+var OpensearchIngest = ComponentDetails{
+	Name:         "os-ingest",
+	EndpointName: "opensearch",
+	//NOTE: update ELASTICSEARCH_WAIT_TARGET_VERSION env (constants.ESWaitTargetVersionEnv) value to match the version reported by this image
```
anders-swanson, line 152: we dont use this var any more (WAIT_TARGET_VERSION

pkg/resources/ingresses/ingress.go
```diff
@@ -149,25 +149,43 @@ func New(vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) ([]*netv1.Ingress, er
 	}
 	if vmo.Spec.Kibana.Enabled {
 		if config.Kibana.OidcProxy != nil {
-			ingresses = append(ingresses, newOidcProxyIngress(vmo, &config.Kibana))
+			ingress := newOidcProxyIngress(vmo, &config.OpenSearchDashboards)
+			ingresses = append(ingresses, ingress)
+			redirectIngress := createRedirectIngressIfNecessary(vmo, existingIngresses, &config.Kibana, &config.OpenSearchDashboardsRedirect)
+			if redirectIngress != nil {
```
anders-swanson, line 155: why is this code not part of createRedirectIngressIfNecessary?

Abhijithang: No compelling reason, I just thought appending the ingresses here would keep the existing flow as it is and easier to understand. Could move annotations to the method to keep it simple?

pkg/config/components.go
Abhijithang: Sure, will remove the comment. Thanks.

pkg/resources/ingresses/ingress.go
anders-swanson: personally i'd put the annotations in the method, since it is repeated multiple places and is a part of creating the ingress object.

Abhijithang: Moving the annotations to the code would require additional parameter, just thought to keep the function as simple as possible. If you insist I could do the change since it does not make much difference either way.

pkg/config/components.go
```diff
@@ -120,6 +145,34 @@ var ElasticsearchIngest = ComponentDetails{
...
+	ReadinessHTTPPath: "/_cluster/health",
+	Privileged:        false,
+	OidcProxy:         &OidcProxy,
+}
+
+// OpensearchIngestRedirect is the default OpensearchIngestRedirect IngestNodes configuration
+var OpensearchIngestRedirect = ComponentDetails{
+	Name:         "os-redirect",
+	EndpointName: "elasticsearch",
+	//NOTE: update ELASTICSEARCH_WAIT_TARGET_VERSION env (constants.ESWaitTargetVersionEnv) value to match the version reported by this image
```
bobdonat, line 166: same comment here as Anders comment above.

anders-swanson (approved):

PR #288, opened 2022-11-25
VZ-7840: Add support for OS dynamic plugin installation

karan56625:
Added coded for OS dynamic plugin installation in OS and OSD

- Introduced **installPlugins** filed in VMI crd for both Opensearch and Opensearch Dashboard.
- Plugin name(if not 3rd party) and plugin url (url to plugin zip file) can be provided in the plugins to install those.
- Supported in both OS and OSD.


pkg/apis/vmcontroller/v1/types.go
```diff
@@ -169,12 +170,18 @@ type (
 		// Minimum count of documents in an index before it is rolled over
 		MinDocCount *int `json:"minDocCount,omitempty"`
 	}
-
 	// Kibana details
 	Kibana struct {
-		Enabled   bool      `json:"enabled" yaml:"enabled"`
-		Resources Resources `json:"resources,omitempty"`
-		Replicas  int32     `json:"replicas,omitempty"`
```
kminder, line 177: We should more broadly review this. 
1. I'm not a huge fan of install in installPlugins
2. What does enabled mean?
3. Should plugins be pluginURLs?

pkg/apis/vmcontroller/v1/types.go
```diff
@@ -169,12 +170,18 @@ type (
...
+		Enabled        bool           `json:"enabled" yaml:"enabled"`
+		Resources      Resources      `json:"resources,omitempty"`
+		Replicas       int32          `json:"replicas,omitempty"`
+		InstallPlugins InstallPlugins `json:"installPlugins,omitempty"`
+	}
+
+	InstallPlugins struct {
+		Enabled bool `json:"enabled" yaml:"enabled"`
+		// List of plugins urls
```
kminder, line 183: Golang convention for acronyms is upper case (i.e. urls -> URLs).

pkg/apis/vmcontroller/v1/types.go
karan56625: 1.  we can consider just `plugins` instead of installPlugins.
2.  Enabled  means whether we want to enable the dynamic installation or not.
3.  Actually plugins could be plugin name, URL to plugins zip file or maven coordinates. May be we can consider to rename it to `installList` 
The complete object will look like similar to what we have in OpenSearch Helm chart.
```
plugins:
  enabled: false
  installList: []
```
Thoughts?

pkg/resources/helper.go
```diff
@@ -53,6 +62,17 @@ const (
...
+     set -euo pipefail
+     # Install OS plugins that are not bundled with OS
+     %s
+    `
+	OSPluginsInstallCmd = `
+    /usr/share/opensearch/bin/opensearch-plugin install -b %s
+	`
+	OSDashboardPluginsInstallCmd = `
+    /usr/share/opensearch-dashboards/bin/opensearch-dashboards-plugin install %s
+	`
```
theabmitra, line 75: Few questions:
1. When are these plugins installed , when pods boot up ?
2. Will this happen for master and data nodes?
3. Also will this happen every time a pod is rebooted ?

karan56625: 1. yes, during pods boot up.
2.  yes, for ingest too.
3.  yes. 
This is similar to what we have in OS helm chart https://github.com/opensearch-project/helm-charts/blob/d18870e3bdccd8fcc8ffb8c5a11da572837ed440/charts/opensearch/values.yaml#L444

pkg/apis/vmcontroller/v1/types.go
karan56625: yes, we can do. But If customer has already define a huge list of plugins and want to disable the plugins for any reason, he doesn't need to remove all the plugins from CR. We can remove it, if it seems redundant. but no harm in putting it there.

pkg/resources/helper.go
theabmitra: This will considerably increase boot time for all the pods . Have we measured the time ?
also please validate this change against the backup-restore pipeline as well. We have a hook that runs after boot, and would want to make sure that dosent break

karan56625: yeah, If there is huge list of plugins, there would be significant increase in boot time. 
yes, I measured the time  with two plugins installation. For single pod, It took around ~70 seconds. And for all OS pods(master, data, ingest), it took around ~4-5minutes.
Will validate the changes against the backup-restore pipeline

pkg/apis/vmcontroller/v1/types.go
kminder: Since this is the syntax used on the OpenSearch Helm chart I think following that convention is fine.

pkg/apis/vmcontroller/v1/types.go
```diff
@@ -169,12 +170,18 @@ type (
 		// Minimum count of documents in an index before it is rolled over
 		MinDocCount *int `json:"minDocCount,omitempty"`
 	}
-
 	// Kibana details
```
kminder, line 173: Can this be changed to OpenSearch Dashboards?

pkg/apis/vmcontroller/v1/types.go
```diff
@@ -169,12 +170,18 @@ type (
...
 	Kibana struct {
 		Enabled   bool      `json:"enabled" yaml:"enabled"`
 		Resources Resources `json:"resources,omitempty"`
 		Replicas  int32     `json:"replicas,omitempty"`
+		Plugins   Plugins   `json:"plugins,omitempty"`
+	}
+
+	Plugins struct {
+		Enabled bool `json:"enabled" yaml:"enabled"`
```
kminder, line 182: Does this need a comment as it could show up in docs?

pkg/resources/helper.go
kminder: Does this really need to be done on all master, data and ingest nodes?

karan56625: It needs to be installed on all nodes. I installed ISM plugin(2.3) on data and master nodes only. But Plugins was not working. After installing it on ingest node, it worked. Still I'll confirm on this, if we really need these plugins on all nodes.

pkg/apis/vmcontroller/v1/types.go
karan56625: done

pkg/apis/vmcontroller/v1/types.go
karan56625: At many places, Kibana and ElasticSearch is getting used instead of OpenSearch and OpenSearch Dashboard. I can rename all of those, but that will be out of this MR scope.

pkg/apis/vmcontroller/v1/types.go
karan56625: Added the comment.

pkg/apis/vmcontroller/v1/types.go
```diff
@@ -169,12 +170,19 @@ type (
...
-
 	// Kibana details
 	Kibana struct {
 		Enabled   bool      `json:"enabled" yaml:"enabled"`
 		Resources Resources `json:"resources,omitempty"`
 		Replicas  int32     `json:"replicas,omitempty"`
+		Plugins   Plugins   `json:"plugins,omitempty"`
+	}
+
+	Plugins struct {
```
kminder, line 181: Should this type be more specific (e.g. OpenSearchPlugins)?

pkg/resources/helper.go
```diff
@@ -497,9 +517,12 @@ func ConvertToRegexp(pattern string) string {
 }
 
 // CreateElasticSearchContainerCMD creates the CMD for OpenSearch containers.
```
kminder, line 519: Fix function name in comment.

pkg/apis/vmcontroller/v1/types.go
karan56625: yeah! But V1Alpha Version of VZ CR still using the ElasticSearch Naming convention. If we making this Plugins more specific, should it be ElasticSearchPlugins for v1alpha1 and OpenSearchPlugins for v1beta1?

karan56625: renamed it to OpenSearchPlugins

pkg/apis/vmcontroller/v1/types.go
```diff
@@ -126,13 +126,14 @@ type (
...
-		Policies   []IndexManagementPolicy `json:"policies,omitempty"`
-		Nodes      []ElasticsearchNode     `json:"nodes,omitempty"`
+		Enabled           bool                    `json:"enabled" yaml:"enabled"`
+		Storage           Storage                 `json:"storage,omitempty"`
+		IngestNode        ElasticsearchNode       `json:"ingestNode,omitempty"`
+		MasterNode        ElasticsearchNode       `json:"masterNode,omitempty"`
+		DataNode          ElasticsearchNode       `json:"dataNode,omitempty"`
+		Policies          []IndexManagementPolicy `json:"policies,omitempty"`
+		Nodes             []ElasticsearchNode     `json:"nodes,omitempty"`
+		OpenSearchPlugins Plugins                 `json:"openSearchPlugins,omitempty"`
```
kminder, line 136: I think you totally missed my point.  I was only suggesting making the type name OpenSearchPlugins.  You change the field name in the struct and the json name in the CRD.  They were fine the way they were.

karan56625: oh My bad! But that type(Plugins) is common for both OpenSearch and OpenSearch Dashboard. Renaming it to OpenSearchPlugins doesn't seem a good option.

karan56625: fixed as discussed.

kminder:
Since there are CRD changes that will end up in the public docs we should also have the doc team review.

kminder (approved):
Approved.

PR #289, opened 2022-11-28
VZ-7845 - Implement retry while updating OS keystore

theabmitra:
Sometimes kubectl exec fails while updating opensearch keystore. This PR will add a retry with a global timeout for the exec operations

### DIFF
--- a/verrazzano-backup-hook/utilities/k8s/k8sHelper.go
+++ b/verrazzano-backup-hook/utilities/k8s/k8sHelper.go
@@ -410,13 +410,13 @@ func (k *K8sImpl) UpdateKeystore(connData *model.ConnectionData, timeout string)
 		return false, err
 	}
 	for _, pod := range esMasterPods.Items {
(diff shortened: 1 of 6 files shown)


verrazzano-backup-hook/utilities/k8s/k8sHelper.go
```diff
@@ -446,3 +448,36 @@ func (k *K8sImpl) UpdateKeystore(connData *model.ConnectionData) (bool, error) {
 	return true, nil
 
 }
+
+func (k *K8sImpl) ExecRetry(pod *v1.Pod, container, timeout string, execCmd []string) error {
```
michael-cico, line 452: We probably need some unit tests for this, if possible

theabmitra: would prefer to implement this in a follow up PR

michael-cico (dismissed):

michael-cico (approved):

PR #297, opened 2022-12-21
VZ-7904: Avoid duplicate index pattern during update index pattern

karan56625:
Duplicate index patterns are created by `updatePatternsInternal` if old system indices are found. 
Add check if saved object already exists with index pattern, avoid updating/creating the same index pattern.


pkg/opensearch_dashboards/opensearch_dashboards_upgrade.go
```diff
@@ -43,7 +43,7 @@ func (od *OSDashboardsClient) updatePatternsInternal(log vzlog.VerrazzanoLogger,
 	}
 	for _, savedObject := range savedObjects {
 		updatedPattern := constructUpdatedPattern(savedObject.Title)
-		if updatedPattern == "" || (savedObject.Title == updatedPattern) {
+		if updatedPattern == "" || isExistingIndexPattern(savedObjects, updatedPattern) {
```
adiforluls, line 46: Can the time complexity be reduced here? Instead of looping through savedObjects slice everytime, if like a map can be introduced for savedObjects and the existence of `updatedPattern` key in that map can be checked?

karan56625: Honestly, I don't want to play around this update logic. Lack of knowledge about the use case of this update logic. It is getting used to convert old style index pattern to new ones. 
Yes, we can increase time complexity here. We can create Map in `getPatterns` function itself. But in that case we can miss the duplicates SavedObject (if they exists). Second, instead of fetching all the saved object, we can fetch the required ones to improve the efficiency. But all these changes require change in update logic itself. 
May be, a new function can be introduced that can creat map from the list and we can check the savedObject title from that map as suggested by you. Although it will contain still one additional for loop, but that will require less changes in the code.

adiforluls: Yeah that's kind of what I had in mind as well, didn't like it too much but that will serve the purpose (i.e. a func that populates a map with savedObjects title as the key before we check for updatedPattern).

adiforluls (approved):

PR #298, opened 2023-01-04
fix deletion of old pattern

karan56625:
In upgrade index-pattern logic, we upgrade the old index-pattern format to new index-pattern format. In few scenario, for some old index-pattern, corresponding new index-pattern already exists. In that case, we were just skipping that scenario and old-index-pattern were persisting. E2E expects these all old index-patterns to get converted into new one. 

In this PR, we delete the old index-pattern for which corresponding new index-patten already exists. For example, if `Verrazzano-namespace-kube-system` and  `verrazzano-namespace-keycloak` exists, Both should be converted into new index pattern `verrazzano-system`. If `verrazzano-system` already exists, we should delete the old index-pattern `Verrazzano-namespace-kube-system` and `verrazzano-namespace-keycloak`


pkg/opensearch_dashboards/opensearch_dashboards_upgrade.go
```diff
@@ -42,16 +42,26 @@ func (od *OSDashboardsClient) updatePatternsInternal(log vzlog.VerrazzanoLogger,
...
+				return fmt.Errorf("failed to delete index pattern %s: %v", savedObject.Title, err)
+			}
 			continue
 		}
 		// Invoke update index pattern API
 		err = od.executeUpdate(log, dashboardsEndPoint, savedObject.ID, savedObject.Title, updatedPattern)
 		if err != nil {
 			return fmt.Errorf("failed to updated index pattern %s: %v", savedObject.Title, err)
 		}
+		savedObject.Title = updatedPattern
```
adiforluls, line 63: Shouldn't `savedObject` be updated before calling the `executeUpdate` func?

pkg/opensearch_dashboards/opensearch_dashboards_upgrade.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2022, Oracle and/or its affiliates.
+// Copyright (C) 2023, 2022, Oracle and/or its affiliates.
```
adiforluls, line 1: Append the new year (i.e. `2022, 2023`)

pkg/opensearch_dashboards/opensearch_dashboards_upgrade.go
adiforluls: Okay looked at the func seems alright

karan56625: Once we have successfully updated the old index pattern format to new one, then only we need to update the Map with new updatePattern. If we do before that and API call gets fails, logic will consider that index pattern to be there in OS.

adiforluls (approved):

PR #299, opened 2023-01-06
VZ-8129: Added Pod Security Standards into OSD

neeraj-n-singh:


pkg/resources/deployments/deployment.go
```diff
@@ -305,6 +305,26 @@ func NewOpenSearchDashboardsDeployment(vmo *vmcontrollerv1.VerrazzanoMonitoringI
 		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.PeriodSeconds = 20
 		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.FailureThreshold = 5
 
+		// When the deployment does not have a pod security context with an FSGroup attribute, any mounted volumes are
+		// initially owned by root/root. The current OSD image creates a group
+		// uid=1000(opensearch-dashboards) gid=1000(opensearch-dashboards) groups=1000(opensearch-dashboards)
+		// with permissons -rw-rw-r--
+		podSecurityContext := corev1.PodSecurityContext{
+			RunAsUser:      resources.New64Val(1000),
```
michael-cico, line 313: Looks good, be sure to update the pod security e2e test so it will check the OSD pods.  Right now they're on an exclude list:

https://github.com/verrazzano/verrazzano/blob/master/tests/e2e/verify-install/security/pod_security_test.go#L40

You can remove that entry and add specific entries for Opensearch, Kiali, and Grafana and run these tests to validate the settings.

michael-cico (approved):

pkg/resources/deployments/deployment.go
anders-swanson: Can you make 1000 a constant please?

anders-swanson: Additionally, this code should be very similar for STS. When STS is done, can you make it a generic code? Since both will need pod security contexts.

neeraj-n-singh: I re-design the code in a bit of generic form.

anders-swanson (approved):

michael-cico (approved):

PR #300, opened 2023-01-06
VZ-8121: Pod security standards for OS

saketmht:


pkg/resources/deployments/deployment.go
```diff
@@ -428,7 +428,7 @@ func getPodSecurityContext(PodUser int64) *corev1.PodSecurityContext {
 	return &corev1.PodSecurityContext{
 		RunAsUser:      &PodUser,
 		FSGroup:        &PodUser,
-		RunAsNonRoot:   resources.NewBool(false),
+		RunAsNonRoot:   resources.NewBool(true),
```
pfmackin, line 431: This should be set at container level since you have different values for the 2 containers.

saketmht: This change is for the OSD pod security which @neeraj-n-singh had merged a couple days ago. He mistakenly set the value to false

michael-cico (approved):

PR #302, opened 2023-01-10
VZ-7933 Remove deprecated Elasticsearch references in VMO CRD Spec

knelson7:


knelson7:
This elastic search updated caused data loss when applying the new crd. Ended up taking another route. Code for this task will be merged in this PR instead (https://github.com/verrazzano/verrazzano-monitoring-operator/pull/306)

PR #303, opened 2023-01-12
VZ-8120: Add default ISM policies

karan56625:
Default ISM policies added.

- **vz-system**
- **vz-application**

These policies are stored in json files under directory `verrazzano-monitoring-operator/k8s/manifests/opensearch`.
By default, VMO reads these policies and create the ISM policies in OpenSearch. 
`spec.components.opensearch.disableDefaultPolicy` is introduced in VMI and VZ CR.  By default, it is false. If `disableDefaultPolicy` is enabled, default policies will be deleted. 

Default ISM policies are immutable. Any update or delete operation on these policies will be reverted immediately.


k8s/manifests/opensearch/vz-system-default-ISM-policy.json
```diff
@@ -0,0 +1,79 @@
...
+
+            }
+          }
+        ]
+      }
+    ],
+    "ism_template":[
+      {
+        "index_patterns":[
+          "verrazzano-system"
```
kminder, line 73: Does this need a * suffix as is done for the default vz-applications policy?

kminder (approved):
Approved.

k8s/manifests/opensearch/vz-system-default-ISM-policy.json
karan56625: `verrazzano-system` is data-stream while `verrazzano-application*` is index pattern that covers all the data-stream created by VZ when we deyploy application e.g. `verrazzano-application-myapp`, `verrazzano-application-bob-book`

PR #304, opened 2023-01-17
VZ-8120: Add default ism policy files in VMO

karan56625:
Add default ism policy files in VMO docker.


kminder:
I don't really like the directory chosen for these (/usr/local/bin/k8s/manifests/opensearch).  Having said that I don't see a different precedent anywhere.  I'll approve this but also ask the team if there is any other precent for a better location.

kminder (approved):
Approved.

docker-images/verrazzano-monitoring-operator/Dockerfile
```diff
@@ -31,12 +31,14 @@ RUN yum update -y \
     && yum install -y openssl \
     && yum clean all \
     && rm -rf /var/cache/yum
-
 COPY --from=build_base /usr/bin/verrazzano-monitoring-operator /usr/local/bin/verrazzano-monitoring-operator
-
+COPY --from=build_base /root/go/src/github.com/verrazzano/verrazzano-monitoring-operator/k8s/manifests/opensearch/vz-application-default-ISM-policy.json /usr/local/bin/k8s/manifests/opensearch/vz-application-default-ISM-policy.json
```
michael-cico, line 35: Is there a reason why this is in `/usr/local/bin` instead of the home dir for the opensearch user, or something along those lines?

karan56625: In the final image, we have only VMO binary file. But now we need few policy files to be read by that binary. VMO is using path from VMO root of these files to read. But in the final image VMO project structure is not there. Only VMO binary is there. If we create these files in home-dir, we need to pass this path to VMO by some env or command line argument. In that case we'll require some default path also for VMO so that it can run locally too.
I was little more inclined to put these policy files in same directory, where the VMO binary exists, made sense to me.

As of now, I am merging this PR as partial code is already merged. Let me know if you have different opinion on putting these files in docker?. I'll raise separate PR for that if there is some concern with the path?

karan56625: PR https://github.com/verrazzano/verrazzano-monitoring-operator/pull/307 created to fix this.

PR #306, opened 2023-01-17
VZ-7933: Add Opensearch and OpensearchDashboards fields to VMI CRD

knelson7:
This pull request adds new fields to the VMI CRD: Opensearch and OpensearchDashboards. The existing fields Elasticsearch and Kibana are tag as deprecated.  In addition, some additional cleanup changes were made related to Elasticsearch and Kibana.


pkg/apis/vmcontroller/v1/types.go
```diff
@@ -137,6 +140,19 @@ type (
 		DisableDefaultPolicy bool                    `json:"disableDefaultPolicy,omitempty"`
 	}
 
+	// Opensearch details
+	Opensearch struct {
```
anders-swanson, line 144: Can you add something to the Elasticsearch type that says it's deprecated?

pkg/vmo/vmospec_test.go
```diff
@@ -51,6 +51,67 @@ func TestInitNode(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		}
+
+		for _, tt := range tests {
+			handleOpensearchConversion(tt.givenVmiSpec)
+			assert.EqualValues(t, tt.expectedVmiSpec.Opensearch.Storage.Size, tt.givenVmiSpec.Opensearch.Storage.Size)
+			assert.EqualValues(t, "", tt.givenVmiSpec.Elasticsearch.Storage.Size)
```
anders-swanson, line 110: should assert that `tt.givenVmiSpec.Elasticsearch == vmcontrollerv1.Elasticsearch{}`

anders-swanson:
Can you also close the old PRs?

anders-swanson (dismissed):

Abhijithang (dismissed):

pkg/vmo/vmospec.go
```diff
@@ -107,6 +113,38 @@ func initNode(node *vmcontrollerv1.ElasticsearchNode, role vmcontrollerv1.NodeRo
...
+		// copy Kibana fields to OpensearchDashboards field
+		spec.OpensearchDashboards = vmcontrollerv1.OpensearchDashboards(*spec.Kibana)
+	}
+
+	// remove old Kibana data
+	spec.Kibana = nil
+}
+
+func handleOpensearchConversion(spec *vmcontrollerv1.VerrazzanoMonitoringInstanceSpec) {
+	if spec.Elasticsearch != nil && spec.Elasticsearch.Enabled {
```
anders-swanson, line 133: what about the case were someone might manually modify the vmi cr with an elasticsearch resource? this would cause both spec.opensearch and spec.elasticsearch to have data, and the existing spec.opensearch would be clobbered by the spec.elasticsearch. Do we want to ignore these kind of changes?

bobdonat: For this usecase, are you saying someone might add the elasticsearch field to an existing CR instead of use the opensearch field.

bobdonat: Usecase does not seem likely.

anders-swanson (approved):

PR #311, opened 2023-01-27
VZ-8222: update to gopkg.in/yaml.v3 v3.0.1

bobdonat:
This pull request is for the update to gopkg.in/yaml.v3 v3.0.1


pkg/config/config.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 
 	"go.uber.org/zap"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
```
mgianatagh, line 10: why is this the only reference updated to yaml.v3?  There are several others in our source.

bobdonat: I don't see any others in VMO

mgianatagh (approved):

PR #315, opened 2023-02-13
VZ-8644:  Reduce VMO image size and cleanup

bobdonat:
This pull request includes the following changes:
1. Reduce the VMO image size from 255MB to 148MB.
2. Use oraclelinux:8-slim to build final image.
3. Changed existing Makefile targets to build without symbol tables. This reduces the image size by about 20MB.
4. Added Makefile debug targets to build VMO with symbol tables if you need to debug.
5. Some cleanup in Makefile

### DIFF
--- a/docker-images/verrazzano-monitoring-operator/Dockerfile
+++ b/docker-images/verrazzano-monitoring-operator/Dockerfile
@@ -30,7 +30,7 @@ FROM ghcr.io/oracle/oraclelinux:8-slim AS final
 
 RUN microdnf update -y \
     && microdnf clean all \
-    && rm -rf /var/cache/yum \
+    && rm -rf /var/cache/yum /var/lib/rpm/__db.* \
     && groupadd -r verrazzano-monitoring-operator \
     && useradd --no-log-init -r -g verrazzano-monitoring-operator -u 1000 verrazzano-monitoring-operator
 
--- a/docker-images/verrazzano-monitoring-operator/Dockerfile
+++ b/docker-images/verrazzano-monitoring-operator/Dockerfile
@@ -20,8 +20,7 @@ ENV GOPATH /root/go
 ENV CGO_ENABLED 0
 COPY . .
 RUN go build \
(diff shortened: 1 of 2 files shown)


docker-images/verrazzano-monitoring-operator/Dockerfile
```diff
@@ -13,28 +13,28 @@ RUN yum update -y \
...
 
-RUN yum update -y \
-    && yum install -y openssl \
-    && yum clean all \
-    && rm -rf /var/cache/yum
-COPY --from=build_base /usr/bin/verrazzano-monitoring-operator /usr/local/bin/verrazzano-monitoring-operator
-WORKDIR /usr/local/bin/
-RUN groupadd -r verrazzano-monitoring-operator && useradd --no-log-init -r -g verrazzano-monitoring-operator -u 1000 verrazzano-monitoring-operator
-RUN chown 1000:verrazzano-monitoring-operator /usr/local/bin/verrazzano-monitoring-operator && chmod 500 /usr/local/bin/verrazzano-monitoring-operator
+RUN microdnf update -y \
```
pshirodkar-orcl, line 31: microdnf when invoked with `--setopt=install_weak_deps=0` does not install unnecessary recommended packages resulting in smaller image. Can you give it a try?

refer https://github.com/cekit/cekit/issues/668. This also mentions about `--setopt=tsflags=nodocs` - I don't the exact impact of this specific flag though.

bobdonat: i will give that a try

bobdonat: This Dockerfile isn't installing anything new - only an update.  Not sure if this matters in this case.

bobdonat: Adding those args did not change anything.

pshirodkar-orcl (approved):

PR #319, opened 2023-02-23
VZ-8707: change master node's pvc owner to VMI

adiforluls:
Setting the master node's PVC `ownerReference` to master node's StatefulSet can cause problems if the Sts is nuked by the controller, the PVCs and PVs will also get nuked. This is a real problem in several cases where upgrade failed because of this issue. This PR is to change the ownerReference of PVCs before the controller can cause accidental deletion of Sts.

### DIFF
--- a/pkg/vmo/statefulset.go
+++ b/pkg/vmo/statefulset.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2020, 2022, Oracle and/or its affiliates.
+// Copyright (C) 2020, 2023, Oracle and/or its affiliates.
 // Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 package vmo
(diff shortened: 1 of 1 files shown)


anders-swanson (approved):

pkg/vmo/statefulset.go
```diff
@@ -152,24 +152,21 @@ func scaleDownStatefulSet(c *Controller, expectedList []*appsv1.StatefulSet, sta
 // to the STS resource, the PVC will automatically get deleted when the STS is deleted.
 // Because PVC is dynamic, when it is deleted, the bound PV will also get deleted.
 // NOTE: This cannot be done automatically using the STS VolumeClaimTemplate.
-func updateOwnerForPVCs(controller *Controller, statefulSet *appsv1.StatefulSet, vmoName string, vmoNamespace string) error {
+func updateOwnerForPVCs(controller *Controller, statefulSet *appsv1.StatefulSet, vmo *vmcontrollerv1.VerrazzanoMonitoringInstance) error {
```
kminder, line 155: Should vmo be vmi?

kminder (approved):
Approved.

pkg/vmo/statefulset.go
adiforluls: It should've been vmi but in the code it's referenced as vmo almost everywhere so wanted to maintain naming consistency.

PR #321, opened 2023-02-23
VZ-8717: Add a delayedLog to reduce error noise

saketmht:


adiforluls (approved):

kminder (dismissed):
Approved.

pkg/vmo/controller.go
```diff
@@ -423,6 +423,7 @@ func (c *Controller) syncHandler(key string) error {
 	if err != nil {
 		zap.S().Errorf("Failed to create controller logger for VMO controller", err)
 	}
+	log.SetFrequency(120)
```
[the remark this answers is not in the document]
saketmht: Yes this will set it for the entire logger. I don't think 2mins is that long though. The reconciliations would still be happening as usual. So in case, the cluster is not healthy it would just not clutter the logs as much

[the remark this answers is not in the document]
saketmht: Updated the PR to use a different logger after some thought.

pkg/vmo/controller.go
```diff
@@ -115,6 +115,10 @@ type Controller struct {
 	// VerrazzanoLogger is used to log
 	log vzlog.VerrazzanoLogger
 
+	// delayedLog is used to log messages at a higher frequency
+	// to reduce error noise in the logs
+	delayedLog vzlog.VerrazzanoLogger
```
gaurav-05, line 120: Have a feeling that if there can be a better name to this variable as both are delayed with different wait.
Also comment says `log messages at a higher frequency` while actually this will produce logs at lower frequency though the delay/wait is higher.

adiforluls: Agree, fix the comment, and maybe `delayed` sends the wrong message. Maybe `lowFrequencyLog`, since this is exactly what this logger is doing.

saketmht: Fixed.

adiforluls (approved):
Approved.

PR #322, opened 2023-02-23
VZ-8763: Modify plugin install logic to fail only the master node in case of install failure.

IshaGirdhar:
Currently, if the plugin installs through vz CR fails, the entire OS cluster goes into a bad state, and due to this any VZ CR update for OS doesn't take place as we have a check where we do not allow any changes to sts/deployments if cluster health is not green. This issue is to enhance the logic to not put the entire cluster into a bad state if plugin installs fail for any reason and only the master node will fail so that the customer can figure out that there was something wrong and at the same time cluster is operational.


kminder (approved):
Approved.

PR #328, opened 2023-02-28
VZ-8550: Create default index patterns in OSD only when it's enabled

adiforluls:


pkg/vmo/controller.go
```diff
@@ -612,10 +612,12 @@ func (c *Controller) syncHandlerStandardMode(vmo *vmcontrollerv1.VerrazzanoMonit
 	/*********************
 	* Add default index patterns
 	**********************/
-	err = c.osDashboardsClient.CreateDefaultIndexPatterns(c.log, resources.GetOpenSearchDashboardsHTTPEndpoint(vmo))
-	if err != nil {
-		c.log.ErrorfThrottled("Failed to add default index patterns : %v", err)
-		errorObserved = true
+	if vmo.Spec.OpensearchDashboards.Enabled != false {
```
karan56625, line 615: should we check for OpenSearch too or OSD can't be enabled if OS is disabled??

adiforluls: There's no such dependency in VPO between OS and OSD. This is the func to instantiate VMI based on VZ CR https://github.com/verrazzano/verrazzano/blob/74e7a19f07c50041ab31ced7d6dfc1a6fce38df1/platform-operator/controllers/verrazzano/component/opensearchdashboards/vmi.go#L21, we just simply look up the effectiveCR to decide whether OSD is enabled or not.

karan56625: then, should we consider the OS check too for index pattern creation. Although OSD check should be enough.

adiforluls: Would OS need to be enabled as well to create default index patterns in OSD? I can find out if you don't know.

karan56625: index patterns are stored in .kibana index

adiforluls: Disabled OS and Enabled OSD, the OSD deployment doesn't come up and the install hangs
```
 Component opensearch-dashboards is waiting for deployment verrazzano-system/vmi-system-osd to exist
```
So OSD won't work without OS. The install would not be successful in this case.

But this bug is specific to when you disable OSD. The install would be successful but this error would be logged continuously in VMO.
```
Failed to add default index patterns : Get \"http://vmi-system-osd.verrazzano-system.svc.cluster.local:5601/api/saved_objects/_find?search=verrazzano-system+or+verrazzano-application\\\\*&type=index-pattern&fields=title&per_page=50&page=1\": dial tcp: lookup vmi-system-osd.verrazzano-system.svc.cluster.local on 10.96.0.10:53: no such host"
```
I don't think we should check OS too, in case OSD is enabled but OS is disabled the install itself wouldn’t succeed.

karan56625: that's what I asked if OSD can be enabled if OS is disabled. If we have handling for that. OSD check is enough.

karan56625: Thanks for verifying that.

karan56625 (approved):

PR #331, opened 2023-03-01
VZ-8729: Enhanced condition to handle to ISM Management

neeraj-n-singh:
This MR will address ISM in opensearch when the customer already has its own customs policy with.

- The system has a custom policy for index vz-system with the same priority and DefaultDisable is not configured.     --> Created missing policy 

- System has a custom policy for index vz-system with higher priority and DefaultDisable is not configured.  ---> Created both default policies.

- System has a custom policy for index vz-system with same priority and DefaultDisable true.      ---> Nothing happened. Single policy exists.

- System has a custom policy for index vz-system with higher priority and DefaultDisable true.     --->Nothing happened. Single policy exists.

- System doesn’t have any policy and DefaultDisable is false.          ---> Default policies created.

- System doesn’t have any policy and DefaultDisable true.    ---> Nothing happened empty system.
Special Case
- System has same name policy with some different conditions, then vz policy will override the same and create ISM policy as per our configuration.

### DIFF
--- a/pkg/opensearch/ism.go
+++ b/pkg/opensearch/ism.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/verrazzano/verrazzano-monitoring-operator/pkg/util/logs/vzlog"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/opensearch/ism.go
```diff
@@ -252,11 +253,7 @@ func (o *OSClient) deletePolicy(opensearchEndpoint, policyName string) (*http.Re
 
 // updateISMPolicyFromFile creates or updates the ISM policy from the given json file.
 // If ISM policy doesn't exist, it will create new. Otherwise, it'll create one.
-func (o *OSClient) updateISMPolicyFromFile(openSearchEndpoint string, policyFileName string, policyName string) (*ISMPolicy, error) {
-	policy, err := getISMPolicyFromFile(policyFileName)
-	if err != nil {
-		return nil, err
-	}
+func (o *OSClient) updateISMPolicyFromFile(openSearchEndpoint string, policyName string, policy *ISMPolicy) (*ISMPolicy, error) {
```
karan56625, line 256: we have extracted the code to read policy from file in this function. Should we consider to rename this to `updateISMPolicy` or something else instead of  `updateISMPolicyFromFile` ?

neeraj-n-singh: Sure. Will update this.

neeraj-n-singh: Done

karan56625 (dismissed):

IshaGirdhar (approved):

PR #336, opened 2023-03-06
JIRA: VZ-8885 update k8s.io libs to 0.25.4

jmaron99:
The k8s.io libs are being updated to the latest patch version for 0.25 - v0.25.4.


THIRD_PARTY_LICENSES.txt
```diff
@@ -2719,5 +2719,5 @@ The above copyright notice and either this complete permission notice or at a mi
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 === ATTRIBUTION-HELPER-GENERATED:
-=== Attribution helper version: {Major:0 Minor:11 GitVersion:0.10.0-102-gd9a1a0a1 GitCommit:d9a1a0a141aea4c18363434b6aef258211712951 GitTreeState:clean BuildDate:2023-02-24T08:51:09Z GoVersion:go1.19.1 Compiler:gc Platform:linux/amd64}
-=== License file based on go.mod with md5 sum:  78b2fe72df4562c78f2b6538987a1705
+=== Attribution helper version: {Major:unknown Minor:unknown GitVersion:unknown GitCommit:unknown GitTreeState:unknown BuildDate:unknown GoVersion:go1.19.2 Compiler:gc Platform:darwin/amd64}
```
fredtibbitts, line 2722: Does it matter that these are "unknown"? Asking because I don't know if it's important to have those filled in correctly.

go.mod
```diff
@@ -37,16 +37,17 @@ require (
...
 
 replace (
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+	github.com/stretchr/testify => github.com/stretchr/testify v1.7.1
 	gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api => k8s.io/api v0.25.2
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.25.2
-	k8s.io/apimachinery => k8s.io/apimachinery v0.25.2
-	k8s.io/client-go => k8s.io/client-go v0.25.2
+	k8s.io/api => k8s.io/api v0.25.4
```
fredtibbitts, line 47: I think these can all be removed, since they are the same versions that are specified in the direct dependencies.

THIRD_PARTY_LICENSES.txt
jmaron99: I'm not certain either - the attribution helper did not complain so I assume it is acceptable

go.mod
jmaron99: My assumption was, and maybe this incorrect, that the replace directives also work on other dependent libs, i.e. version 0.25.4 would be replaced in other libs specifying version 0.25.2.  Otherwise I wasn't certain what purpose the "replace" section serves?

fredtibbitts: I believe you would only need to specify the replacements if there are indirect dependencies on these modules that pull in different versions from the direct dependencies, but that does not seem to be the case here. I'm guessing the replacements were needed at one point and we never removed them. It's easy enough to check, just remove them and run `go mod tidy` and if it results in no additional changes, you should be good.

THIRD_PARTY_LICENSES.txt
jmaron99: Downloaded and run go install.

This is the sort of checkin/PR that drives me crazy - simple change requiring an unwieldy amount of review and possible changes.

go.mod
jmaron99: No additional changes in go.mod?  There are plenty of changes in go.sum

THIRD_PARTY_LICENSES.txt
jmaron99: So I'm hoping we don't spend an inordinate amount of time on this...

go.mod
fredtibbitts: go.sum changes are fine, if nothing changed in go.mod then you don't need the replacements.

go.mod
```diff
@@ -37,16 +37,17 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.2.0 // indirect
 	gopkg.in/ini.v1 v1.62.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	k8s.io/component-base v0.25.2 // indirect
+	k8s.io/component-base v0.25.4 // indirect
 )
 
 replace (
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+	github.com/stretchr/testify => github.com/stretchr/testify v1.7.1
```
fredtibbitts, line 45: Sorry I didn't catch this the first time, but you upgraded testify from 1.7.1 to 1.8.0 in the direct dependencies, but then this replace is downgrading it to 1.7.1, so the upgrade in the direct dependencies is a no-op. Is that what you intended?

jmaron99: Yes - unless we want to pause and get a PLS approved for the new testify version?

PR #339, opened 2023-04-05
VZ-9210: Add annotation to OS/OSD pods to add hooks to delay application startup until the pod proxy is ready to accept traffic

karan56625:
Added annotations `proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'` to OS/OSD pods.
This annotation adds hooks to delay application startup until the pod proxy is ready to accept traffic, mitigating some startup race conditions.

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -65,15 +65,21 @@ const (
 	`
 	OSMasterPluginsInstallTmpl = `
      set -euo pipefail
+     # wait for Sidecar proxy to be ready
+     until curl -sS --head localhost:15021/healthz/ready; do echo Waiting for Sidecar; sleep 3 ; done ;
      # Install OS plugins that are not bundled with OS
      %s
     `
 	OSIngestPluginsInstallTmpl = `
      set +euo pipefail
+     # wait for Sidecar proxy to be ready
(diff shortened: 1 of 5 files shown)


pkg/resources/helper.go
```diff
@@ -65,15 +65,21 @@ const (
 	`
 	OSMasterPluginsInstallTmpl = `
      set -euo pipefail
+     # wait for Sidecar proxy to be ready
+     until curl -sS --head localhost:15021/healthz/ready; do echo Waiting for Sidecar; sleep 3 ; done ;
```
kminder, line 69: This needs to handle installs where Istio is disabled.

pfmackin, line 69: Please add the name of the container that you are waiting for.
You can also try adding  pod annotation to bypass envoy outbound port 443

karan56625: I can try that. would that be alright if bypass outbound port 443 for OS/OSD?? If we do, this script change also not required

karan56625: added annotation `"holdApplicationUntilProxyStarts": true` to delay application startup until the isitio pod proxy is ready

karan56625: instead of change in start up script, now added annotation to delay application startup until the pod proxy is ready.

kminder (approved):
Approved.

PR #347, opened 2023-04-21
VZ-9459.  Explicitly annotate ingresses with the Verrazzano cluster issuer

michael-cico:
Adds the annotation

```
	ingress.Annotations["cert-manager.io/cluster-issuer"] = "verrazzano-cluster-issuer"
```

to the ingresses managed by the VMO.  This is to allow integration with externally installed Cert-Manager instances where the default ClusterIssuer is not the Verrazzano issuer.

### DIFF
--- a/pkg/resources/ingresses/ingress.go
+++ b/pkg/resources/ingresses/ingress.go
@@ -18,6 +18,8 @@ import (
 
 var defaultIngressClassName = "verrazzano-nginx"
 
+const verrazzanoClusterIssuerName = "verrazzano-cluster-issuer"
+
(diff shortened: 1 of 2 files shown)


pkg/resources/ingresses/ingress.go
```diff
@@ -89,6 +91,7 @@ func createIngressElementNoBasicAuth(vmo *vmcontrollerv1.VerrazzanoMonitoringIns
 	}
 
 	ingress.Annotations["cert-manager.io/common-name"] = hostName
+	ingress.Annotations["cert-manager.io/cluster-issuer"] = verrazzanoClusterIssuerName
```
kminder, line 94: Needs copyright update
Needs unit tests

michael-cico: updated with new unit tests

kminder (approved):
Approved

PR #354, opened 2023-06-22
VZ-10009: Changes to default ISM Policy

saketmht:
Changes to default ISM policy:

- Removed cold state from ISM policies
- Transition to delete state, 14 days after rollover
- System and application ISM policy are now same

Other changes:

- When default ISM policy is enabled/created/updated, attach the policy to active write index of system and application data streams if it was previously being managed by our default policy or no policy at all
- When default ISM policy is disabled, remove the policy from active write index of data streams.

### DIFF
--- a/pkg/opensearch/ism.go
+++ b/pkg/opensearch/ism.go
@@ -300,13 +300,15 @@ func (o *OSClient) createOrUpdateDefaultISMPolicy(log vzlog.VerrazzanoLogger, op
 			if err != nil {
 				return defaultPolicies, err
 			}
-			defaultPolicies = append(defaultPolicies, createdPolicy)
 			// When the default policy is created or updated
 			// Add default policy to the current write index of the data stream
-			err = o.addDefaultPolicyToDataStream(log, openSearchEndpoint, createdPolicy)
-			if err != nil {
-				return defaultPolicies, err
+			if createdPolicy != nil {
+				err = o.addDefaultPolicyToDataStream(log, openSearchEndpoint, createdPolicy)
+				if err != nil {
+					return defaultPolicies, err
+				}
 			}
+			defaultPolicies = append(defaultPolicies, createdPolicy)
 		}
 	}
 	return defaultPolicies, nil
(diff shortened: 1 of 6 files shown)


pkg/opensearch/ism.go
```diff
@@ -282,12 +301,167 @@ func (o *OSClient) createOrUpdateDefaultISMPolicy(log vzlog.VerrazzanoLogger, op
...
+		writeIndices, err := o.getWriteIndexForDataStream(log, openSearchEndpoint, pattern)
+		if err != nil {
+			return fmt.Errorf("failed to get the write index for %s: %v", pattern, err)
+		}
+
+		for _, index := range writeIndices {
+			// Check if the index is currently being managed by our default policy
+			// If not then don't add the default policy
+			ok, err := o.isManagedByDefaultPolicy(openSearchEndpoint, index, policyName)
+			if err != nil {
```
gaurav-05, line 337: Are we applying only if the index had a default policy ? If yes it will not fix the issue of not applying policy to existing indexes.

saketmht: The policy is being applied if the index had our policy attached to it or no policy at all

saketmht: I will change the function name to something like `shouldAttachDefaultPolicy` to clarify that.

pkg/opensearch/ism.go
```diff
@@ -18,6 +18,24 @@ import (
 )
 
 type (
+	DataStreams struct {
+		DataStream []DataStream `json:"data_streams,omitempty"`
```
adiforluls, line 22: `DataStreams` and should the json be `dataStreams`?

pkg/opensearch/ism.go
```diff
@@ -18,6 +18,24 @@ import (
...
+		Indices        []Index   `json:"indices,omitempty"`
+		Generation     int       `json:"generation,omitempty"`
+		Status         string    `json:"status,omitempty"`
+		Template       string    `json:"template,omitempty"`
+	}
+	TimeStamp struct {
+		Name string `json:"name,omitempty"`
+	}
+	Index struct {
+		Name string `json:"index_name,omitempty"`
```
adiforluls, line 36: Should the json names be `indexName` and `indexUUID`?

pkg/opensearch/ism.go
saketmht: The json are formatted according to the OS api response since this is used to decode the response from OS requests.

pkg/opensearch/ism.go
saketmht: Same comment as above.

IshaGirdhar (approved):

PR #363, opened 2023-07-07
VZ-10223: Update pod startup command to check if plugin is already installed

saketmht:


pkg/resources/helper.go
```diff
@@ -63,29 +63,62 @@ const (
 	echo "Commenting out java heap settings in jvm.options..."
 	sed -i -e '/^-Xms/s/^/#/g' -e '/^-Xmx/s/^/#/g' config/jvm.options
 	`
+	PluginInstalledCheckWithExit = `
+     exit_status=$?
+     if [ $exit_status -ne 0 ]; then
+          if grep -iq "plugin .* already exists" /tmp/error.log; then
+               echo "Plugin already installed. Skipping installation."
```
adiforluls, line 70: Should you be cleaning temp directory here?

pkg/resources/helper.go
```diff
@@ -63,29 +63,62 @@ const (
...
+               cat /tmp/error.log
+               exit 1
+          fi
+     fi
+    `
+	PluginInstalledCheck = `
+     exit_status=$?
+     if [ $exit_status -ne 0 ]; then
+          if grep -iq "plugin .* already exists" /tmp/error.log; then
+               echo "Plugin already installed. Skipping installation."
```
adiforluls, line 81: Same question about temp directory clean up.

pkg/resources/helper.go
```diff
@@ -63,29 +63,62 @@ const (
...
+          else
+               cat /tmp/error.log
+          fi
+     fi
+    `
+	DashboardsPluginInstalledCheck = `
+     exit_status=$?
+     if [ $exit_status -ne 0 ]; then
+          if grep -iq "plugin .* already exists" /tmp/error.log; then
+               echo "Plugin already installed. Skipping installation"
```
adiforluls, line 91: Same question about temp directory cleanup.

pkg/resources/helper.go
saketmht: Added a line below to clean up the temp file being created

pkg/resources/helper.go
saketmht: Updated

pkg/resources/helper.go
saketmht: Updated

adiforluls (approved):

PR #389, opened 2023-09-15
VZ-11000: Refactor backup hook to work with Opster Opensearch

saketmht:


verrazzano-backup-hook/constants/constants.go
```diff
@@ -73,14 +83,22 @@ const (
 	// IngestLabelSelector Opensearch ingest pod label selector
 	IngestLabelSelector = "app=system-es-ingest"
 
+	NewIngestStatefulSetName = "opensearch-es-ingest"
```
gaurav-05, line 86: Using New prefix at all places does not seem to be right, can we implement some struct or better naming

saketmht: I am not sure what struct would achieve here. And for naming I didn't want to use Opster so just added New prefix. Any other suggestion?

verrazzano-backup-hook/opensearch/interface.go
```diff
@@ -58,15 +65,97 @@ type OpensearchImpl struct {
...
+	OperatorDeploymentLabelSelector string
+	// OpenSearchMasterLabel Label selector for OpenSearch master pods
+	OpenSearchMasterLabel string
+	// OpenSearchDataLabel Label selector for OpenSearch data pods
+	OpenSearchDataLabel string
+}
+
+func NewOpensearchVar(isLegacyOS bool) *OpensearchVar {
+	if isLegacyOS {
+		return &OpensearchVar{
```
gaurav-05, line 130: Does it make sense to remove declaration from constants and keep it here only to be consistent for both conditions.

saketmht: The constants are being used in tests as well. So removing them would just mean additional refactoring

gaurav-05: Ok

IshaGirdhar (dismissed):

PR #395, opened 2023-09-26
disable security plugin for OS/OSD (#387)

karan56625:
* disable security plugin for OS/OSD


pkg/resources/helper.go
```diff
@@ -39,7 +39,7 @@ const (
     `
 	OpenSearchDashboardCmdTmpl = `#!/usr/bin/env bash -e
     %s
-	/usr/local/bin/opensearch-dashboards-docker`
+	./opensearch-dashboards-docker-entrypoint.sh`
```
lake-of-dreams, line 42: does the image in 1.5 contains this path?

PR #396, opened 2023-09-26
disable security plugin for OS/OSD (#387)

karan56625:
* disable security plugin for OS/OSD

### DIFF
--- a/pkg/resources/helper.go
+++ b/pkg/resources/helper.go
@@ -37,7 +37,7 @@ const (
 	/usr/local/bin/docker-entrypoint.sh`
 	OpenSearchDashboardCmdTmpl = `#!/usr/bin/env bash -e
     %s
-	/usr/local/bin/opensearch-dashboards-docker`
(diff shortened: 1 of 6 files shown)


pkg/resources/helper.go
```diff
@@ -37,7 +37,7 @@ const (
 	/usr/local/bin/docker-entrypoint.sh`
 	OpenSearchDashboardCmdTmpl = `#!/usr/bin/env bash -e
     %s
-	/usr/local/bin/opensearch-dashboards-docker`
+	./opensearch-dashboards-docker-entrypoint.sh`
```
lake-of-dreams, line 40: does image in 1.4  contain this path?

gaurav-05: @karan56625 I think this should be needed 1.5 + only you can cross check

karan56625: yep, closing this one. We just need to update for 1.6 and 1.5 only. 1.4 contains OS 1.2.3.

PR #405, opened 2023-11-23
VZ-11405: Fix Golang version from 1.19 to 1.20.10

SmritiOjha23:


docker-images/verrazzano-monitoring-operator/Dockerfile
```diff
@@ -31,6 +20,7 @@ RUN groupadd -r verrazzano-monitoring-operator && \
     useradd --no-log-init -r -g verrazzano-monitoring-operator -u 1000 verrazzano-monitoring-operator
 
 FROM $FINAL_IMAGE AS final
+FROM ghcr.io/oracle/oraclelinux:8-slim
```
lake-of-dreams, line 23: probably not required as you already have `FROM $FINAL_IMAGE AS final`

PR #415, opened 2024-01-08
VZ-11405: Updating to 1.20.10 for release 1.7

SmritiOjha23:


THIRD_PARTY_LICENSES.txt
```diff
@@ -15,6 +15,8 @@ Copyright (C) 2022, Oracle and/or its affiliates.
 Copyright (c) 2020 Oracle America, Inc. and its affiliates.
 Copyright (c) 2020, 2021, Oracle and/or its affiliates.
 Copyright (c) 2020, 2022, Oracle and/or its affiliates.
+Copyright (c) 2020, 2023, Oracle and/or its affiliates.
+Copyright (c) 2022, 2023, Oracle and/or its affiliates.
```
karan56625, line 19: shouldn't we add copyright for 2024 too?

docker-images/verrazzano-monitoring-operator/Dockerfile
```diff
@@ -1,18 +1,14 @@
...
 
-RUN yum update -y \
-    && yum-config-manager --save --setopt=ol7_ociyum_config.skip_if_unavailable=true \
-    && yum install -y oracle-golang-release-el7 \
-    && yum-config-manager --disable ol7_developer_golang\* \
-    && yum-config-manager --enable ol7_developer_golang119 \
-    && yum -y install golang \
-    && yum clean all \
-    && go version
+FROM ghcr.io/oracle/oraclelinux:8-slim AS build_base
```
karan56625, line 5: can't we use Golang image here instead of installing Golang.

SmritiOjha23: Hi Karan, when I raised my PR for master, I was using the Golang image for 1.20.10 only but Bhatt commented "Go 1.20.10 is now available via yum repository for OL8. Although we have built few components using the custom Golang image, for the new components you can also install golang-1.20.10-1 from OEL8 base image.
https://corparch-core-srv.slack.com/archives/GEPMXPFTK/p1700596017761479" and so I changed this in this way.

